PR #1, opened 2022-02-15
First commit of GUI of BLE app.

JL-Money:
This is a crude outline of the GUI component of the BLE telemetry app. It has a bottom navigation bar.

PR #2, opened 2022-02-22
Parse C

MichaelCampos-eng:
This is a program that parses C code. 

Input: string of C code

eg.  Parse test = new Parse(code);

Output: Object that holds a repository of all C variables and its values.


src/parsingCopy/src/com/company/Parse.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    /**Accesses all the nodes of the tree*/
+    private void saveNodes(IASTNode node) throws ExpansionOverlapsBoundaryException {
+        IASTNode[] children = node.getChildren();
+        for (IASTNode child :children) {
+            if (child instanceof CPPASTSimpleDeclaration) {
+
+                /**Setup*/
+                CPPASTSimpleDeclaration declaration = (CPPASTSimpleDeclaration) child;
```
ducky64, line 35: Style nit: it may be helpful to add an example here of what is being extracted, eg `uint8_t PEDAL_POS = 8;` here, at least if these are different (I'm not sure)

src/parsingCopy/src/com/company/Parse.java
```diff
@@ -0,0 +1,163 @@
...
+                CPPASTNamedTypeSpecifier specifier = (CPPASTNamedTypeSpecifier) declaration.getDeclSpecifier();
+                CPPASTDeclarator declarator = (CPPASTDeclarator) declaration.getDeclarators()[0];
+                CPPASTEqualsInitializer init = (CPPASTEqualsInitializer) declarator.getInitializer();
+
+                /**Get what is needed: Value, Name, bigType & smallType*/
+                CPPASTLiteralExpression value = (CPPASTLiteralExpression) init.getInitializerClause(); //get value
```
ducky64, line 41: Style comment (... about comments!): comments should generally not be redundant with what can be understood by reading the code - in this case, I would say that these short comments are unnecessary, since you did a good job of writing self-documenting code.

In general, comments are often useful for providing examples, summarizing behavior, or documenting tricky behavior / oddities.

src/parsingCopy/src/com/company/Parse.java
```diff
@@ -0,0 +1,163 @@
...
+                CPPASTName name = (CPPASTName) declarator.getName(); //get name
+                IToken bigType = specifier.getSyntax(); //get big type (iToken)
```
ducky64, line 43: What do you mean big type and small type? From the below, if you mean bigType for things like `const`, it's more formally called a [type qualifier](https://en.cppreference.com/w/cpp/language/cv), while `uint32_t` is the actual type. Using (and finding) standardized terminology is helpful for writing clear code.

There might also be other ways to simplify this distinction - we should talk about it.

src/parsingCopy/src/com/company/Parse.java
```diff
@@ -0,0 +1,163 @@
...
+
+                /**Saves name and value to their respective location*/
+                HashMap<CPPASTName, CPPASTLiteralExpression> contentMap = getContentMap(bigType, smallType);
+                contentMap.put(name, value);
+            }
+        }
+    }
+
+    /**Adds a new hashmap for Large Types if bType does not exist*/
+    private void checkBigTypeExists(IToken bType) {
```
ducky64, line 58: The function name (check...) implies this should return something, but it doesn't? Perhaps you mean "recordBigType" or "addBigType"?

We might also think of ways to restructure the data structures to simplify this - we should discuss that.

Similar for below.

src/parsingCopy/src/com/company/Parse.java
```diff
@@ -0,0 +1,163 @@
...
+import org.eclipse.cdt.core.parser.*;
+import org.eclipse.cdt.internal.core.dom.parser.cpp.*;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Parse {
+
+    private IASTTranslationUnit translationUnit; //current translation unit node
+    public HashMap<IToken, HashMap> bigTypes;
```
ducky64, line 17: Did you know you can (I think) also add type parameters for the inner hash map, eg `HashMap<IToken, HashMap<..., ...>>`? Comprehensively documenting types is very useful.

src/parsingCopy/src/com/company/Parse.java
```diff
@@ -0,0 +1,163 @@
...
+
+    public Parse(String code) throws Exception {
+        bigTypes = new HashMap<>();
+
+        translationUnit = getIASTTranslationUnit(code.toCharArray());
+        IASTNode node = translationUnit;
```
ducky64, line 23: Is this line needed? Can you pass `translationUnit` directly into `saveNodes`?

src/parsingCopy/src/com/company/Parse.java
```diff
@@ -0,0 +1,163 @@
...
+    private void checkBigTypeExists(IToken bType) {
+        if (!bigTypes.containsKey(bType)) {
+            HashMap<CPPASTName, HashMap> subTypes = new HashMap();
+            bigTypes.put(bType, subTypes);
+        }
+    }
+
+    /**Adds a new hashmap for Small Types if sType does not exist*/
+    private void checkSmallTypeExists(IToken bType, CPPASTName sType) {
+        checkBigTypeExists(bType);
```
ducky64, line 67: Does it make sense to have checkSmallType and checkBigType as separate functions, since they are always called together, and one is invalid without the other? It's a lot of "season-to-taste", the more self-contained functions are, the better.

src/parsingCopy/src/com/company/Parse.java
```diff
@@ -0,0 +1,163 @@
...
+    /**Only works if sType and bTYpe exist*/
+    private HashMap<CPPASTName, CPPASTLiteralExpression> getContentMap(IToken bType, CPPASTName sType) {
+        HashMap subTypes = bigTypes.get(bType);
+        HashMap contents = (HashMap<CPPASTName, CPPASTLiteralExpression>) subTypes.get(sType);
+        return contents;
+    }
+
+    /**Gets value of given name*/
+    public CPPASTLiteralExpression get(CPPASTName name) {
+        //Iterate over big types
```
ducky64, line 84: If you find yourself iterating over a hash map for a lookup operation, that's a bit of a red flag and might indicate you may want a different data structure. Let's discuss this too.

src/parsingCopy/src/com/company/Parse.java
```diff
@@ -0,0 +1,163 @@
...
+                }
+
+            }
+
+        }
+        throw new IllegalArgumentException("Given name does not exist!");
+    }
+
+    /**Initializes the translation unit setup*/
+    public static IASTTranslationUnit getIASTTranslationUnit(char[] code) throws Exception {
```
ducky64, line 108: Can `code` be a string instead?

src/parsingCopy/src/com/company/Parse.java
```diff
@@ -0,0 +1,163 @@
...
+        return GPPLanguage.getDefault().getASTTranslationUnit(fc, si, ifcp, (IIndex)idx, options, log);
+    }
+
+    /** Below are the functions used to print all*/
+
+    /**Function called by instance to print all types*/
+    public void showNameValue() {
+        iterateBigTypes();
+    }
+    /**Iterate through all big types eg. const*/
```
ducky64, line 126: In production code, I generally prefer not to leave in debugging code. Obviously a bit of season-to-taste (especially if you're building data structures utilities).

If you directly print the HashMap, will it do the right thing and print the contents?

src/parsingCopy/src/com/company/ParseTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+        Parse test = new Parse(code);
+
+        String value1 = "0x16";
+        String value2 = "0x32";
+        String value3 = "0x64";
+
+        String n1 = "CAN_ID";
+        String n2 = "CAN_ID2";
+        String n3 = "CAN_ID3";
```
ducky64, line 26: Feel free to inline these into the below lines for conciseness.

Sometimes, declaring values separately can be a stylistic benefit, by giving them a name, but this test code would be clear with the "magic numbers" (or magic strings in this case) directly inline.

src/parsingCopy/src/com/company/ParseTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+        CPPASTName name1 = new CPPASTName(n1.toCharArray());
+        CPPASTName name2 = new CPPASTName(n2.toCharArray());
+        CPPASTName name3 = new CPPASTName(n3.toCharArray());
+
+        String litExpr1 = test.get(name1).getRawSignature();
```
ducky64, line 32: Would it make sense to have `get` take a string as its argument instead of this `CPPASTName`?

src/parsingCopy/src/com/company/ParseTest.java
```diff
@@ -0,0 +1,107 @@
...
+        String litExpr3 = test.get(name3).getRawSignature();
+
+        Assert.assertEquals(value1, litExpr1);
+        Assert.assertEquals(value2, litExpr2);
+        Assert.assertEquals(value3, litExpr3);
+    }
+
+    @Test
+    public void checkLitExprdiverse() throws Exception {
+        String code = "const uint16_t CAN_ID = 0x16; \n" +
```
ducky64, line 43: Similar comments here - I believe inlining should make this test a lot more readable.

ducky64 (commented):
Looked at Parse.java and ParseTest.java - both are on the right track and seem functionally fine, though there's some style comments (general note: this is always expected from a code review, and style is important for code readability and maintainability in team projects) and some larger points we can discuss.

Additionally, there seems to be a lot of other irrelevant files - can you do a first pass on purging those? Good practice is to keep the repository clean and pull requests focused on the feature at hand. Pulling in external code directly may also have copyright implications.

ducky64:
I'll also acknowledge that part of this is probably a bit chaotic (with regards to file organization, not the particular code written) since I didn't set up a template project so it's less clear what should go where. We should also work through that, which might also be related to #1.

PR #3, opened 2022-02-27
Concise Parser

MichaelCampos-eng:
Removed other unnecessary branches and made code cleaner. Functions the same but better.


src/com/company/Parse.java
```diff
@@ -0,0 +1,120 @@
...
+                    continue;
+                }
+
+                if (specifier == null) {
+                    continue;
+                }
+
+                if (declaration == null) {
+                    continue;
+                }
```
ducky64: You should be able to consolidate all these into one `if` statement, by OR-ing all of them (eg, `if (init == null || declarator == null || ...) {`)

src/com/company/Parse.java
```diff
@@ -0,0 +1,120 @@
...
+import org.eclipse.cdt.core.parser.*;
+import org.eclipse.cdt.internal.core.dom.parser.cpp.*;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+
+public class Parse {
+
+    private IASTTranslationUnit translationUnit; //current translation unit node
```
ducky64: This actually isn't used anywhere else, can you get rid of it? (and only keep `translationUnit` scoped in the constructor?)
Keeping variables scoped as narrowly as possible eases code reading and understanding, since it's clear that it can't be accessed (or worse, modified) elsewhere.

src/com/company/Parse.java
```diff
@@ -0,0 +1,120 @@
...
+    public HashMap<String, Contents> repo;
```
ducky64: Will this work?
```suggestion
    public HashMap<String, Contents> repo = new HashMap<>();
```
(and delete the initialization in the constructor)

If the value doesn't depend on any constructor argument, we might as well initialize it on declaration.

src/com/company/Parse.java
```diff
@@ -0,0 +1,120 @@
...
+                Contents contents = new Contents(name, value, typeQualifier, type);
+                repo.put(name.getRawSignature(), contents);
+            }
+        }
+    }
+
+    public Contents get(String key) throws IllegalAccessError{
+        if (repo.get(key) == null) {
+            throw new IllegalAccessError("The name does not exist!");
+        }
```
ducky64: It might be cleaner to use an assertion here, though that needs some project-wide configs. Let's get #1 merged first, then see if we can work an assertion in here.

Actually, it may be better for this to return `null` - especially since that behavior would be consistent with how `HashMap` handles things, which will be what Java developers will be used to.

src/com/company/Parse.java
```diff
@@ -0,0 +1,120 @@
...
+                if (contents.typeQualifer.equals(key)) {
+                    list.add(contents);
+                }
+            }
+        }
+        if (list.isEmpty()) {
+            throw new IllegalArgumentException("No contents[s] were found for given key.");
+        }
+        return list;
+    }
```
ducky64: What's the use case for this? If it's just to enable testing, consider moving this into the test code as a utility function.

src/com/company/Parse.java
```diff
@@ -0,0 +1,120 @@
...
+    public ArrayList<Contents> getByType(String key) {
+        return findList(key, 0);
+    }
+
+    public ArrayList<Contents> getTypeQualifier(String key) {
+        return findList(key, 1);
+    }
+
+    /**Makes and return a list of contents*/
+    private ArrayList<Contents> findList (String key, int i) {
```
ducky64: Consider making this a higher-order function.
Or, alternatively, just allow users to get the underlying HashMap `repo` and iterate over it themselves.

src/com/company/ParseTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class ParseTest {
+
+    @Test
+    public void doesNotExist() throws Exception {
```
ducky64: It might be useful to separate the "doesn't exist" case from the "malformed code" case. On line 15, that's an example of doesn't exist, while lines 16-17 likely won't compile or possibly even won't parse.

It would also be useful to check a case where the variable name isn't even present in the code at all.

src/com/company/Repository.java
```diff
@@ -0,0 +1,18 @@
+package com.company;
```
ducky64: What's this for? I don't see it used anywhere.

src/com/company/Utils.java
```diff
@@ -0,0 +1,240 @@
+package com.company;
```
ducky64: What's this for? I don't think it's being used anywhere? Can we get rid of it?

The code rights also aren't clear - I see an external author on it, but it's doesn't have a license attached.

src/com/company/serialException.java
```diff
@@ -0,0 +1,20 @@
+package com.company;
```
ducky64: Similar - if this isn't being used, can we get rid of it?

ducky64 (changes requested):
A few things to clean up, but overall looking good!

Once #1 is merged and we have a project structure, we should make sure this code goes in the right folders and is structurally compatible. We might also want to see if we can fetch CDT as a dependency (via Gradle - part of the default build system setup for the template Android project) instead of pulling in the JAR directly.

MichaelCampos-eng:
The struct feature has been added and passes local test.

app/src/main/java/com/example/bottomnav/Contents.java
```diff
@@ -0,0 +1,23 @@
+package com.example.bottomnav;
+
+import org.eclipse.cdt.core.parser.IToken;
+import org.eclipse.cdt.internal.core.dom.parser.cpp.CPPASTLiteralExpression;
+import org.eclipse.cdt.internal.core.dom.parser.cpp.CPPASTName;
+
+import java.util.HashMap;
+
+public class Contents {
+
```
ducky64: Try to be consistent with style - I typically don't like having extraneous whitespace lines that don't serve a purpose (such as visually separating different chunks). But as with anything style, YMMV.

In the larger picture, we might want to adopt a style guide guide and set up an auto-formatter, but that's a future thing.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -0,0 +1,99 @@
...
+import org.eclipse.cdt.core.parser.*;
+import org.eclipse.cdt.internal.core.dom.parser.cpp.*;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+
+public class Parse {
+    public HashMap<String, Contents> constRepo = new HashMap<>();
+    public HashMap<String, ArrayList> structRepo = new HashMap<>();
```
ducky64: Can you also define the type parameter of ArrayList, for example `HashMap<String, ArrayList<something>>`?

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -0,0 +1,99 @@
...
+    private ArrayList subStrucCon = null;
+
+    public Parse(String code) throws Exception {
+        IASTTranslationUnit translationUnit = getIASTTranslationUnit(code.toCharArray());
+        IASTNode node = translationUnit;
+        saveNodes(node);
```
ducky64: Will this work instead?
```suggestion
        saveNodes(translationUnit);
```

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /**Accesses all the nodes of the tree*/
+    private void saveNodes(IASTNode node) throws Exception {
```
ducky64: A more descriptive name might be better here. I think `process` or `traverse` (esp for trees) is more common terminology for what you're doing here, `save` more implies filesystem access or doing some kind of backup operation.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -0,0 +1,99 @@
...
+        IASTNode[] children = node.getChildren();
+        for (IASTNode child : children) {
+
+            if (child instanceof CPPASTSimpleDeclaration) {
```
ducky64, line 24: Can you use a [combined instanceof and cast](https://www.baeldung.com/java-pattern-matching-instanceof)? (needs a somewhat recent version of Java, not sure what we're on)

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -0,0 +1,99 @@
...
+                    structHelper(compSpec.getDeclarations(false), name.getRawSignature());
+                } else {
+                    saveAs("const", child);
+                }
+
+            }
+        }
+    }
+
+    private void saveAs(String t, IASTNode node) throws Exception{
```
ducky64: This behaves pretty different in the `const` initializer and `struct` case. I'd suggest making them two different functions.

If you have common code that is used by both, you might refactor that into a helper function that is called by both, for example `getTypeOfDeclaration` or something, to avoid the code duplication that might otherwise happen with two functions.

Note, in some software engineering guides, boolean flags (which `t` here is effectively equivalent to) are sometimes treated with suspicion, because it's a sign that your function might be doing two different things.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -0,0 +1,99 @@
...
+                    constRepo.put(name.getRawSignature(), contents);
+                }
+
+            }
+
+        }
+    }
+
+    private void structHelper(IASTDeclaration[] declarations, String name) throws Exception {
+        subStrucCon = new ArrayList();
```
ducky64: I'd suggest minimizing the amount of state kept in instance variables where not necessary. Here, `subStructCon` seems mainly being used to share data between `structHelper` and `saveAs`, instead of being useful persistent state. Instead, consider having `saveAs` (or the refactored version of that) return either the filled ArrayList, or have it return an element to add to an ArrayList created solely within `stuctHelper`

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -0,0 +1,99 @@
...
+                    Contents contents = new Contents(name, value, typeQualifier, type);
+                    constRepo.put(name.getRawSignature(), contents);
+                }
+
+            }
+
+        }
+    }
+
+    private void structHelper(IASTDeclaration[] declarations, String name) throws Exception {
```
ducky64: A more descriptive name could be helpful here, like `extractStructMembers` or something. `helper` is pretty ambiguous.

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -0,0 +1,171 @@
+package com.example.bottomnav;
+
+import org.junit.Test;
+import static org.junit.Assert.*;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+
+
+public class ParseTest {
```
ducky64, line 8: Consider having two tests for the const and struct case? The behavior is quite different, unified only by that they're both parsed from the same C source file.

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -0,0 +1,171 @@
...
+        assertEquals("uint16_t", test.get("CAN_ID2").type);
+        assertEquals("uint16_t", test.get("CAN_ID3").type);
+
+        assertEquals("0x16", test.get("CAN_ID").value);
+        assertEquals("0x32", test.get("CAN_ID2").value);
+        assertEquals("0x64", test.get("CAN_ID3").value);
+    }
+
+    @Test
+    public void checkdiverse() throws Exception {
```
ducky64: Consider a more specific / descriptive name that gets at what this test is checking for - perhaps it's the different types?

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -0,0 +1,171 @@
...
+                "  uint8_t reserved2;\n" +
+                "  uint8_t reserved3;\n" +
+                "};";
+        Parse test = new Parse(code);
+        HashMap<String, ArrayList> structRepo = test.structRepo;
+
+        assertEquals(true, structRepo.containsKey("ChargerControlStruct"));
+        assertEquals(false, structRepo.containsKey("struct ChargerControlStruct"));
+
+        ArrayList<StructContents> contents = structRepo.get("ChargerControlStruct");
```
ducky64: The `containsKey` checks above are kind of redundant with this?

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+        Parse test = new Parse(code);
+
+        HashMap<String, Contents> repository = test.constRepo;
+
+        assertEquals(true, repository.containsKey("CAN_ID1"));
+        assertEquals(true, repository.containsKey("CAP_ID2"));
+        assertEquals(true, repository.containsKey("CAN_ID3"));
+
+        assertEquals("const", test.get("CAN_ID1").typeQualifer);
```
ducky64: The `containsKey` checks above are kind of redundant with this?

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -0,0 +1,171 @@
...
+        assertEquals("uint32_t", test.get("CAP_ID2").type);
+        assertEquals("uint32_t", test.get("CAN_ID3").type);
+
+        assertEquals("0x16", test.get("CAN_ID1").value);
+        assertEquals("0x18", test.get("CAP_ID2").value);
+        assertEquals("0x64", test.get("CAN_ID3").value);
+    }
+
+    @Test
+    public void listCheck() throws Exception {
```
ducky64: This seems like it's more testing the `MakeList` utility class? What's `MakeList` for? It seems like its only purpose is to be tested, which is kind of odd?

ducky64 (changes requested):
Overall looks fine, but some suggestions for code quality in the inline comments.

app/src/main/java/com/example/bottomnav/Contents.java
ducky64: Bump

app/src/main/java/com/example/bottomnav/Parse.java
ducky64: Bump

app/src/test/java/com/example/bottomnav/ParseTest.java
ducky64: bump

app/src/test/java/com/example/bottomnav/ParseTest.java
ducky64: bump - recommend deleting the containsKey checks

app/src/test/java/com/example/bottomnav/ParseTest.java
ducky64: bump - see above

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -0,0 +1,108 @@
...
+                    saveStuct(compSpec.getDeclarations(false), name.getRawSignature());
+                } else {
+                    saveConst(child);
+                }
+
+            }
+        }
+    }
+
+    private ArrayList getComponents(IASTNode node) throws Exception {
```
ducky64: May be helpful to have documentation for what this does, and the assumptions of the inputs (here, the input needs to be a CPPASTSimpleDeclaration). Really, what this seems to be doing is trying to parse a CPPASTSimpleDeclaration into a name, type, type-qualifier, and initializer.

Also highly recommend having another struct-like type that saves the type of the stuff you're packing into the output, instead of an ArrayList takes any Object (and only knows it as Objects). Having items indexed by a name is also cleaner than having them indexed by a number, especially since each slot in the ArrayList has a different meaning which is count to typical usage.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -0,0 +1,108 @@
...
+            IToken typeQualifier = (IToken) components.get(1);
+            CPPASTName type = (CPPASTName) components.get(2);
+            CPPASTEqualsInitializer init = (CPPASTEqualsInitializer) components.get(3);
+            CPPASTLiteralExpression value = (CPPASTLiteralExpression) init.getInitializerClause();
+            Contents contents = new Contents(name, value, typeQualifier, type);
+            constRepo.put(name.getRawSignature(), contents);
+        }
+    }
+
+    private void saveStrucMem(IASTNode node, ArrayList struct) throws Exception {
```
ducky64: Consider putting this directly into saveStruct? This isn't called anywhere else, and I don't think this give you any significant abstraction / readability benefits?

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    private ArrayList getComponents(IASTNode node) throws Exception {
+        ArrayList components = new ArrayList();
+        CPPASTSimpleDeclaration declaration = (CPPASTSimpleDeclaration) node;
+        CPPASTNamedTypeSpecifier specifier = (CPPASTNamedTypeSpecifier) declaration.getDeclSpecifier();
+        CPPASTDeclarator declarator = (CPPASTDeclarator) declaration.getDeclarators()[0];
+        CPPASTEqualsInitializer init = (CPPASTEqualsInitializer) declarator.getInitializer();
+
+        if (declarator != null || specifier != null || declaration != null && init != null) {
```
ducky64: Is this the correct logic? Under what cases should things be allowed to be null?

ducky64 (commented):
Progress! Some more comments though.

ducky64 (commented):
A step in the right direction, but think about taking it a bit further:
- The code currently creates an object, and then checks if it's valid. There's a lot of logic going on in the `Components` constructor, but what about instead making the object simple and putting the parsing code elsewhere?
- And if you do the parsing and validation before the constructor, you may not ever have to call the constructor and get an invalid object - that is, you have the guarantee that all objects are valid (the parsing and validation may return empty / null though. I'll put an example below.
- Use of optional: think of it as a way to better communicate that a value may be `null`. That everything can be null is one of the gotchas of Java, and Optional is a way to explicitly indicate that something can be empty within the type system (as opposed to having it implicit that something could be null, and hoping that the programmer is sharp enough to catch it)
  - Optional types are more common in languages and programming practices that lean more functional

Example
```java
public class Components {
  public string name;
  public string typeQualifier;
  public string type;
  public CPPASTEqualsInitializer init;

  // very simple constructor, arguments directly become the new object's fields
  // note that there's no valid field - the constructor can't fail
  public Components(String inName, String inTypeQualifier, ...) {
    name = inName;
    typeQualifier = inTypeQualifier;
    ...
  }

  // called with Components.Deconstruct(decl), but because this is not a constructor
  // this can return a different type, like Optional.empty or null
  public static Optional<Components> Deconstruct(CPPASTSimpleDeclaration decl) {
    CPPASTNamedTypeSpecifier specifier = (CPPASTNamedTypeSpecifier) declaration.getDeclSpecifier();
    CPPASTDeclarator declarator = (CPPASTDeclarator) declaration.getDeclarators()[0];
    init = (CPPASTEqualsInitializer) declarator.getInitializer();
    if (declarator != null && specifier != null && declaration != null) {
      // you might need to do additional validation here or unpacking to String here
      // you might also have init be Optional since not every declaration has an init field
      return Optional.of(new Components(
        (CPPASTName) declarator.getName(),
        (IToken) specifier.getSyntax(),
        (CPPASTName) specifier.getName(),
        (CPPASTEqualsInitializer) declarator.getInitializer()
      ));
    } else {
      return Optional.empty();
    }
  }
}
```

Does this make a bit more sense?

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -0,0 +1,81 @@
...
+                    storeConst(declaration);
+                }
+            }
+        }
+    }
+
+    private void storeConst(CPPASTSimpleDeclaration declaration) throws Exception {
+        Optional data = Components.Deconstruct(declaration);
+        if (data.isPresent() && (((Components) data.get()).init.isPresent())) {
+            Components comp = (Components) data.get();
```
ducky64: Will this work? (can you type parameterize the Optional so that `.get()` knows it's a Component out?
```suggestion
        Optional<Components> data = Components.Deconstruct(declaration);
        if (data.isPresent() && ((data.get().init.isPresent())) {
            Components comp = data.get();
```

You might even just use `data.get()` instead of `comp` if you have one or two usages.

Also, similarly below - I highly recommend type parameterizing where you have the option so the compiler can do more work for you!

ducky64 (commented):
One more change, and I think this is good to go.

ducky64 (approved):

PR #4, opened 2022-04-10
text parsed

MichaelCampos-eng:
There is now a new folder called resources located in test folder. The ParseTest file has a new junit test that converts an input stream to a ByteArrayOutputStream. After calling Parse on the toString of output stream, it verifies three const's and two struct's. Assuming it works for now, but can check every single line in future.


app/src/main/java/com/example/bottomnav/OpenTextFile.java
```diff
@@ -0,0 +1,26 @@
+package com.example.bottomnav;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Optional;
+public class OpenTextFile {
+    Optional<char[]> code;
+
+    public OpenTextFile(String fileName) throws IOException {
```
ducky64: I don't think you want a class here - this is more of a function that would return a string (or char array).

In Java, since everything needs to be in a class, you'd make this a static method - that is, a function of the class (instead of, of an instance). So you'd never instantiate the class, the class only exists as a container for the function.

It's up to you whether you want to make it return Optional or just exception out. I'd probably just make it exception out, since the usual assumption is that this will succeed.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -16,17 +22,28 @@ public class Parse {
     public HashMap<String, Contents> constRepo = new HashMap<>();
     public HashMap<String, ArrayList<StructContents>> structRepo = new HashMap<>();
 
-    public Parse(String code) throws Exception {
-        IASTTranslationUnit translationUnit = getIASTTranslationUnit(code.toCharArray());
+    public static Optional<Parse> parseTextFile(String fileName) throws Exception {
```
ducky64: Similarly, I'd probably just make this throw an exception if the code file failed to open. If you change OpenTextFile to throw an exception, you can just call it directly and assume success (and if not successful, OpenTextFile will throw the exception for you).

app/src/main/res/values-w600dp/dimens.xml
```diff
@@ -0,0 +1,3 @@
+<resources>
+    <dimen name="fab_margin">48dp</dimen>
+</resources>
```
ducky64: What's up with the changes to the Android resources? Consider backing out these changes if they're not relevant - otherwise it might create a conflict if someone else is working on the same file?

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -126,4 +124,56 @@ public void structNConst() throws Exception {
         assertEquals("uint16_t", test.getConst("CAN_ID3").type);
         assertEquals("0x64", test.getConst("CAN_ID3").value);
     }
+
+    @Test
+    public void parseData() throws Exception {
+        Optional<Parse> open = Parse.parseTextFile("parseData.txt");
```
ducky64: Recommend naming the file with a `.h` extension for a C++ header file.
It shouldn't affect your code, but is a tad more realistic than `.txt`.

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -126,4 +124,56 @@ public void structNConst() throws Exception {
...
+        Parse test = open.get();
+        HashMap<String, Contents> repository = test.constRepo;
+        ArrayList<StructContents> struct1 = test.getStruct("ChargerControlStruct");
+        ArrayList<StructContents> struct2 = test.getStruct("ChargerControlStruct");
```
ducky64: Is there a reason this is duplicated?

ducky64 (commented):
Mostly looking good! A few recommended code quality fixes.

app/src/main/java/com/example/bottomnav/OpenTextFile.java
MichaelCampos-eng: When I try to use it in the static method I get this error: "Non-static method 'getClass()' cannot be referenced from a static context" when I try to get an Input string.

MichaelCampos-eng: Nevermind :)

ducky64 (approved):
lgtm

PR #5, opened 2022-05-25
Mapped ID to Struct

MichaelCampos-eng:
The current parser does not read comments. I think the translation unit ignores comments, so I basically compared names of const variables to struct names. It works well when const variables have structure name in consecutive order.


app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -174,4 +173,25 @@ public void parseData() throws Exception {
         assertEquals("uint8_t", struct.get(4).type);
         assertEquals("uint8_t", struct.get(5).type);
     }
+
+    @Test
+    public void checkIDToStructMap() throws Exception {
+        Parse test = Parse.parseTextFile("parseData.h");
+        ArrayList<StructContents> struct1 = test.getAssociatedStruct("chargerstatus");
```
ducky64: Consider doing more of an end-to-end test - instead of providing the stylized input, can you pass in the CAN ID name as you'd get in actual usage?

app/src/test/resources/parseData.h
```diff
@@ -132,8 +133,13 @@ const uint16_t CAN_TRITIUM_MEASURE_BASE_L = 0x440;
 const uint16_t CAN_TRITIUM_COMMAND_BASE_R = 0x420;
 const uint16_t CAN_TRITIUM_MEASURE_BASE_R = 0x460;
 
+// Struct
```
ducky64: What's up with these?

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -71,6 +100,17 @@ private void storeStuct(IASTDeclaration[] declarations, String name) throws Exce
...
 
+    public ArrayList getAssociatedStruct(String IDName) {
+        for (String struct : IDStruct.keySet()) {
+            PayloadMap curr = IDStruct.get(struct);
+            System.out.println("This is ID: " + IDName + " and Struct: " + struct);
+            if (curr.canIDNames.contains(IDName)) {
+                return getStruct(curr.struct);
+            }
+        }
+        return null;
```
ducky64: Can this whole thing be simplified? Instead of storing the `simplifyName` result for each CAN ID const in a dedicated data structure and searching through the entire structure, can you just run `simplifyName` on `IDName` and index into the IDStruct from there?

Note that iterating through a list is relatively expensive (since it needs to look for every element), compared to a hash table lookup (when you know exactly what key you're looking for). For this application, it practically won't matter (since the lists are small enough), but it's contrary to good practice.

Additionally, if you can simplify this, you might be able to get rid of PayloadMap completely, where `idStruct` is a direct simplified name to struct map.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -71,6 +100,17 @@ private void storeStuct(IASTDeclaration[] declarations, String name) throws Exce
         structRepo.put(name, struct);
     }
 
+    public ArrayList getAssociatedStruct(String IDName) {
```
ducky64: Style: convention for variable names tend to be camelCase, starting with lower case - so use `idName` here - and similarly `idStruct` above

ducky64 (commented):
Cool! If parsing comments is truly not supported, this looks reasonable. Some suggested improvements in inline comments.

Though, for comments: I did some quick searching and found [a thread on CDT comment handling](https://www.eclipse.org/forums/index.php/t/53932/) - it seems like there might be a flag you pass in to get it to create comment nodes? Reason is that comments are explicit, compared to name matching which is implicit and the structure might not be obvious to someone making changes to these files - so if an explicit solution is possible it would be preferred.

Process note: it would be helpful in the PR description to have a bit more detailed summary of the PR and approach used - for example here, that you simplify both CAN IDs and names to a common representation by converting everything to lowercase and removing `Can` (do we need to remove `Can`? - or is this an inconsistency in can_id.h that we can fix?) and `Struct`. It might also help to have that as comments somewhere in the code itself. Examples are also helpful for describing things.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -14,11 +14,13 @@
 import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 public class Parse {
     public HashMap<String, ConstContents> constRepo = new HashMap<>();
     public HashMap<String, ArrayList<StructContents>> structRepo = new HashMap<>();
+    public HashMap<String, PayloadMap> idStruct = new HashMap<>();
```
ducky64: Are you sure you can't get away with `HashMap<String, ArrayList<StructContents>>` and that you need a PayloadMap data structure?

ducky64 (commented):
Can you make the association syntax a bit cleaner? For example, perhaps:
```c++
// struct_def CAN_PEDAL_POS = CanPedalPosStruct
```

or even better, if it can be a comment associated with the CAN ID, for example:
```c++
const uint16_t CAN_PEDAL_POS = ...;  // struct_def CanPedalPosStruct
```

(but this assumes that you can get comments associated with an AST node, which I don't know if the parser provides)

In general, while the symbols get the job done (and can be very concise in terms of characters), they're not very readable. Also feel free to iterate on the syntax, this is what I could come up with in 2 minutes.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -71,6 +76,51 @@ private void storeStuct(IASTDeclaration[] declarations, String name) throws Exce
         structRepo.put(name, struct);
     }
 
+    public void mapPayload(IASTNode[] comments) {
```
ducky64: We should discuss the parsing algorithm, and how we can make it work with the syntax we'd like instead of fitting the syntax to an easy parser.

I also recommend learning regex for when you need to do text parsing - try https://regexone.com/
Regex can give you powerful parsing rules without the complexity of implementing it from scratch. Java should have regex support.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -71,6 +76,51 @@ private void storeStuct(IASTDeclaration[] declarations, String name) throws Exce
...
+        while (character != '|') {
+            name.append(character);
+            start += 1;
+            character = line[start];
+        }
+        return name.toString();
+    }
+
+    //Given a CAN ID name, retrieves its associated struct
+    public ArrayList getAssociatedStruct(String idName) {
```
ducky64: Why does this return an `ArrayList`? Is there ever a case where you can return more than one struct for a given name? Would `Optional<StructContents>` be a better fit?

Similarly for `getStruct` - maybe that needs to be refactored?

ducky64 (commented):
Few more comments and suggestions looking through the code.

Good progress, though!

MichaelCampos-eng:
Ready for review

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -63,7 +63,7 @@ private void storeConst(CPPASTSimpleDeclaration declaration) throws Exception {
     }
 
     private void storeStuct(IASTDeclaration[] declarations, String name) throws Exception {
-        ArrayList<StructContents> struct = new ArrayList();
+        ArrayList<StructContents> struct = new ArrayList<>();
```
ducky64: Can you get away with this? since you already define the type parameter in the type declaration
```suggestion
        ArrayList<StructContents> struct = new ArrayList);
```

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -74,48 +74,29 @@ private void storeStuct(IASTDeclaration[] declarations, String name) throws Exce
...
-                int secondIndex = idName.length() + 2;
-                String structName = findString(line, secondIndex);
-                if (!idStruct.containsKey(structName)) {
-                    PayloadMap rep = new PayloadMap(structName);
-                    idStruct.put(structName, rep);
+            Matcher matcher = pattern.matcher(comment.getRawSignature());
+            if (matcher.find()) {
+                if (!idStruct.containsKey(matcher.group(2))) {
+                    ArrayList<String> list = new ArrayList<>();
+                    idStruct.put(matcher.group(2), list);
```
ducky64:
```suggestion
                    idStruct.put(matcher.group(2), new ArrayList<>());
```

I don't think having the intermediary `list` is too useful?

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -9,18 +9,19 @@
...
 import java.util.Map;
 import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 public class Parse {
     public HashMap<String, ConstContents> constRepo = new HashMap<>();
     public HashMap<String, ArrayList<StructContents>> structRepo = new HashMap<>();
-    public HashMap<String, PayloadMap> idStruct = new HashMap<>();
+    public HashMap<String, ArrayList<String>> idStruct = new HashMap<>();
```
ducky64: Since there are two `String`s in the type parameters, I recommend having a comment that defines the meaning of each string - for example `// CAN ID -> List<CAN Struct>`

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -74,48 +74,29 @@ private void storeStuct(IASTDeclaration[] declarations, String name) throws Exce
...
-
-    //Given a CAN ID name, retrieves its associated struct
-    public ArrayList getAssociatedStruct(String idName) {
+    //Given a CAN ID name, retrieves its associated struct contents
+    public ArrayList<StructContents> getAssociatedStruct(String idName) {
         for (String struct : idStruct.keySet()) {
-            PayloadMap curr = idStruct.get(struct);
-            if (curr.canIDNames.contains(idName)) {
-                return getStruct(curr.struct);
+            ArrayList<String> curr = idStruct.get(struct);
```
ducky64: If you find yourself iterating over a hashmap to find a particular value, you may be doing your data structures wrong

ducky64 (commented):
Looking better, but a few more lingering issues

MichaelCampos-eng:
Ready for review. I basically added a new hashMap (canIDToStruct) that maps a CAN ID to a struct. There is another hash map (structToCanIds) that remains which maps a can struct to an array of can IDs. This might be useful if the user needs to access all can IDs associated to a particular struct without having to iterate over canIDToStruct.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -69,18 +77,35 @@ private void storeStuct(IASTDeclaration[] declarations, String name) throws Exce
             }
         }
         structRepo.put(name, struct);
+
```
ducky64: stray newline?
```suggestion
```

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -174,4 +173,26 @@ public void parseData() throws Exception {
...
+        ArrayList<StructContents> struct3 = test.getCanStruct("CAN_STRAIN_DATA");
+        ArrayList<StructContents> struct4 = test.getCanStruct("CAN_PEDAL_POS");
+
+        //Retrieve contents of structure from PayLoadMap object
+        ArrayList<StructContents> stru1 = test.getStruct("ChargerStatusStruct");
+        ArrayList<StructContents> stru2 = test.getStruct("ChargerControlStruct");
+        ArrayList<StructContents> stru3 = test.getStruct("CanStrainGaugeStruct");
+        ArrayList<StructContents> stru4 = test.getStruct("CanPedalPosStruct");
+
+        assertEquals(stru1, struct1);
```
ducky64: Can you inline these? `struct1` and `stru1` don't really add anything in terms of readability (and similarly for below)
```suggestion
        assertEquals(test.getStruct("ChargerStatusStruct"), test.getCanStruct("CAN_CHARGER_STATUS"));
```

ducky64 (commented):
Just a few more stylistic issues, otherwise looks good!

ducky64 (approved):
lgtm if unit tests pass

PR #7, opened 2022-06-08
Decode

MichaelCampos-eng:
I created a new file called decode.h that contains the small test cases provided. Basically, this program now reads comments from an .h file and modifies contents of a CAN ID, specifically its `payLoadDataType` variable. This allows the function `decode` to read the type of the payload, single or struct, given a can id number.  If it's a single, the can id number is assigned to a decoded float in the `payloadmap` hash map. I wasn't sure what the key value should be in that case. All other cases link variables of a struct to decoded values. So as of right now, the hash map takes in ambiguous keys and values which I find myself needing to cast returned values. Probably I should create a object which contains multiple hash maps for different types of data. Also in ParseTest, certain bytes needed to be casted which I don't know why. New changes also consisted of renaming old functions in order to make them easier to read.


app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -23,8 +29,12 @@ public class Parse {
     public HashMap<String, ConstContents> constRepo = new HashMap<>();
     // Struct name -> StructContents
     public HashMap<String, ArrayList<StructContents>> structRepo = new HashMap<>();
-    // CAN ID -> CAN Struct
-    public HashMap<String, String> IDToStruct = new HashMap<>();
+    // CAN ID Name -> CAN Struct
+    public HashMap<String, String> canStructRepo = new HashMap<>();
```
ducky64: Consider naming this closer to its function, eg `CanIdToStructName`. This could also obliviate the comment above - self documenting code!
(and similarly for the other HashMaps you have)

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -23,8 +29,12 @@ public class Parse {
...
+    // CAN ID -> CAN ID Name
+    public HashMap<Integer, String> canIdRepo = new HashMap<>();
+    // Variable name or CAN ID -> decoded value
+    public HashMap payloadMap = new HashMap();
```
ducky64: Even if Java doesn't enforce it, you should always include the type parameter - so both you and the reader know what's going on here.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -50,7 +60,39 @@ public Parse(char[] code) throws Exception {
                 }
             }
         }
-        associateIDStruct(comments);
+        parseComments(comments);
+    }
+
+    @RequiresApi(api = Build.VERSION_CODES.O)
```
ducky64: What's this for? Is there a way around it that doesn't require Android-specific code and instead sticks to standard Java APIs?

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -50,7 +60,39 @@ public Parse(char[] code) throws Exception {
...
+    public void decode(int canID, byte[] payload) {
+        String payloadDataType = getConst(canID).payLoadDataType;
+        switch (payloadDataType){
```
ducky64: This would be a great place to make payloadDataType a struct.

Using inheritance / a class hierarchy may also be a good idea here. We can discuss that if you're interested.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -50,7 +60,39 @@ public Parse(char[] code) throws Exception {
                 }
             }
         }
-        associateIDStruct(comments);
+        parseComments(comments);
+    }
+
+    @RequiresApi(api = Build.VERSION_CODES.O)
+    public void decode(int canID, byte[] payload) {
```
ducky64: Should this just return the result instead of stuffing it into a HashMap? The HashMap seems pretty roundabout?

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -50,7 +60,39 @@ public Parse(char[] code) throws Exception {
...
+        String payloadDataType = getConst(canID).payLoadDataType;
+        switch (payloadDataType){
+            case "single":
+                payloadMap.put(canID,
+                        ByteBuffer.wrap(payload).order(ByteOrder.LITTLE_ENDIAN).getFloat());
```
ducky64: What about the case where it's not a float? For example, what if it's a single int?

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -50,7 +60,39 @@ public Parse(char[] code) throws Exception {
...
+                break;
+            case "struct":
+                ArrayList<StructContents> structContents = getStruct(canID);
+                int incr = payload.length / structContents.size();
```
ducky64: You shouldn't treat structs as repeating - once you parse an entire struct you should discard the rest of the data (if any).

If you wanted to be really robust and thorough, you can generate an error type (or error annotation) if there is leftover data. Similarly, think about how to handle when there isn't enough data.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -50,7 +60,39 @@ public Parse(char[] code) throws Exception {
...
+                    } else {
+                        payloadMap.put(structContents.get(i).name,
+                                ByteBuffer.wrap(byteArray).order(ByteOrder.LITTLE_ENDIAN).getFloat());
+                    }
+                }
+                break;
+        }
+    }
+
+    private byte[] getBytes(byte[] payload, int start, int num) {
```
ducky64: Generally, if some functionality is a basic data structure operation, it's probably part of the standard library and just a StackOverflow / Google search away. I'd include taking a slice of an array into that category. See https://stackoverflow.com/questions/11001720/get-only-part-of-an-array-in-java - reinventing the wheel is not recommended if it can be avoided.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -50,7 +60,39 @@ public Parse(char[] code) throws Exception {
...
+            case "struct":
+                ArrayList<StructContents> structContents = getStruct(canID);
+                int incr = payload.length / structContents.size();
+                for (int i = 0; i < structContents.size(); i++) {
+                    byte[] byteArray = getBytes(payload, i * incr, incr);
+                    if (incr == 1) {
+                        payloadMap.put(structContents.get(i).name, byteArray[0] & 0xff);
+                    } else {
+                        payloadMap.put(structContents.get(i).name,
+                                ByteBuffer.wrap(byteArray).order(ByteOrder.LITTLE_ENDIAN).getFloat());
```
ducky64: I'm not sure this length detection logic is sound. Let's discuss ways to make this more robust and generalizable.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -61,6 +103,11 @@ private void storeConst(CPPASTSimpleDeclaration declaration) throws Exception {
                     (CPPASTLiteralExpression) comp.init.get().getInitializerClause();
             ConstContents contents = new ConstContents(comp.name, value, comp.typeQualifier,
                     comp.type);
+            Pattern pattern = Pattern.compile("\\dx(\\d+)");
```
ducky64: Is there any way to get the value as an int out of the AST? Or some way for the AST to evaluate the literal value? The more the AST / C++ parsing libraries can do for us, the better - for example, if the value is `42` or `0x2a` - or maybe even `20+22`.

But if you do want to match, it should be as strict as possible. Hex numbers in C start with `0x`, you shouldn't match on any-digit-then-x.

Also, consider moving this into the ConstContents, perhaps having a ConstContents.decode static method that can reject bad values.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -79,28 +126,42 @@ private void storeStuct(IASTDeclaration[] declarations, String name) throws Exce
...
         for (IASTNode comment : comments) {
-            Matcher matcher = pattern.matcher(comment.getRawSignature());
-            if (matcher.find()) {
-                IDToStruct.put(matcher.group(1), matcher.group(2));
+            Matcher matcher1 = pattern1.matcher(comment.getRawSignature());
+            Matcher matcher2 = pattern2.matcher(comment.getRawSignature());
+            if (matcher1.find()) { // associate ID to struct
+                canStructRepo.put(matcher1.group(1), matcher1.group(2));
+            } else if (matcher2.find()) { // assoicate payload data type
+                getConst(matcher2.group(1)).payLoadDataType = matcher2.group(2);
```
ducky64: Can you merge your internal format for these two cases? Both are variants of "this CAN ID has a payload of some format", whether that format is a single data type (eg, uint32_t or float) or a struct. We can discuss how to architect this cleanly.

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -187,4 +189,26 @@ public void checkIDToStructMap() throws Exception {
         assertEquals(test.getStruct("CanPedalPosStruct"),
                 test.getCanStruct("CAN_PEDAL_POS"));
     }
+
+    @Test
+    public void decodingSimple() throws Exception {
+        Parse test = Parse.parseTextFile("decode.h");
+        byte[] packedFloatPayload = {0x71, (byte) 0xFD, 0x47, 0x41};
```
ducky64, line 205: Can you place the test data with the test code?

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -187,4 +189,26 @@ public void checkIDToStructMap() throws Exception {
...
+        byte[] twovarPayload = {0x0B, 0x00, 0x00, (byte) 0xA5};
+        byte[] fourVarPayload = {(byte) 0xED, (byte) 0xC7, 0x00, 0x43, (byte) 0xD1, 0x53, 0x23, 0x41};
```
ducky64: What's up with the random `(byte)` conversions?

app/src/test/resources/decode.h
```diff
@@ -0,0 +1,22 @@
...
+  uint8_t reserved1Pos;
+  uint8_t reserved2Pos;
+};
+
+// @canPayloadStruct CAN_TRITIUM_VELOCITY = CanTritiumVelocityStruct
+// @payloadDataType CAN_TRITIUM_VELOCITY = struct
+const uint16_t CAN_TRITIUM_VELOCITY = 0x402;
+
+struct CanTritiumVelocityStruct {
+  uint8_t rpm;
```
ducky64: wait I'm pretty sure these are supposed to be floats

ducky64 (changes requested):
Good start, but I think this needs to handle additional cases. Let's find some time to discuss the different data types this should handle, and ideas on architecting the system to support those.

app/src/main/java/com/example/bottomnav/Components.java
```diff
@@ -22,12 +25,51 @@ public Components(CPPASTName givenName, IToken givenTQ,
...
+            case NamedType:
+                return deconstruct((CPPASTNamedTypeSpecifier) dec.getDeclSpecifier(),
+                        (CPPASTDeclarator) dec.getDeclarators()[0]);
+
+            default:
+                return Optional.empty();
+        }
+    }
+
+    private static CPPASTType getSpecifierType(CPPASTSimpleDeclaration declaration) {
```
ducky64: Since you only use CPPASTType once here, I'd instead suggest inlining this function into where it's used (above) and getting rid of the enum in favor of comments.

If you had to decide between string and enum, go for the enum, but in this case I think you can skip the problem entirely and have less code.

app/src/main/java/com/example/bottomnav/Components.java
```diff
@@ -22,12 +25,51 @@ public Components(CPPASTName givenName, IToken givenTQ,
         init = givenInit;
     }
 
-    public static Optional<Components> Deconstruct(CPPASTSimpleDeclaration declaration) throws Exception {
-        CPPASTNamedTypeSpecifier specifier = (CPPASTNamedTypeSpecifier) declaration.getDeclSpecifier();
-        CPPASTDeclarator declarator = (CPPASTDeclarator) declaration.getDeclarators()[0];
+    public static Optional<Components> getComponents(CPPASTSimpleDeclaration dec) throws Exception {
```
ducky64, line 24: Would be useful to have a comment on what this does, and perhaps some example input/output pairs

app/src/main/java/com/example/bottomnav/Components.java
```diff
@@ -22,12 +25,51 @@ public Components(CPPASTName givenName, IToken givenTQ,
...
+        CPPASTType specifierType = getSpecifierType(dec);
+        switch (specifierType) {
+            case Simple:
+                return deconstruct((CPPASTSimpleDeclSpecifier) dec.getDeclSpecifier(),
```
ducky64: Also consider inlining these functions.

app/src/main/java/com/example/bottomnav/ConstContents.java
```diff
@@ -4,17 +4,19 @@
 import org.eclipse.cdt.internal.core.dom.parser.cpp.CPPASTLiteralExpression;
 import org.eclipse.cdt.internal.core.dom.parser.cpp.CPPASTName;
 
-import java.util.HashMap;
 public class ConstContents {
     String value;
     String name;
     String typeQualifer;
     String type;
+    PayLoadDataType payLoadDataType;
```
ducky64: Is this redundant with `String type` above?

app/src/main/java/com/example/bottomnav/ConstContents.java
```diff
@@ -4,17 +4,19 @@
...
 
     public ConstContents(CPPASTName iName, CPPASTLiteralExpression iValue, IToken iTQ, CPPASTName iType) throws Exception {
         value = iValue.getRawSignature();
         typeQualifer = iTQ.toString();
         name = iName.getRawSignature();
         type = iType.getRawSignature();
     }
+
+
```
ducky64: Style: get rid of extraneous newlines?

app/src/main/java/com/example/bottomnav/DecodedData.java
```diff
@@ -0,0 +1,33 @@
...
+            case Double:
+                return new DecodedDouble();
+            case Long:
+                return new DecodedLong();
+            case uint8_t:
+                return new DecodedInteger();
+            case uint16_t:
+                return new DecodedInteger();
+            default:
+                return null;
```
ducky64: I'd recommend using an Optional here, so it's clear when things fail. Lurking nulls can be a debugging nightmare, whereas having an Optional forces the function user to handle the error case.

app/src/main/java/com/example/bottomnav/PayLoadDataType.java
```diff
@@ -0,0 +1,6 @@
+package com.example.bottomnav;
+
+// May have to add more types in the future
+public enum PayLoadDataType {
```
ducky64: What is this needed for? Isn't this kind of redundant with the DecodedData class hierarchy?

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -13,8 +13,8 @@ public void doesNotExist() throws Exception {
         String code = "const uint16_t CAN_ID;";
         Parse test = new Parse(code.toCharArray());
 
-        assertEquals(null, test.getConst("CAN_ID"));
-        assertEquals(null, test.getConst("ducky"));
+        assertEquals(null, test.getConstContents("CAN_ID"));
+        assertEquals(null, test.getConstContents("ducky"));
```
ducky64: 🦆

app/src/main/java/com/example/bottomnav/Translation.java
```diff
@@ -0,0 +1,43 @@
+package com.example.bottomnav;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+
+public class Translation {
+    public static byte[] payload;
```
ducky64: Using a global variable to pass data around is a very bad idea, because it's super hard to reason about - for example, what writes to it? what is reading from it? how are the two synchronized?

It's much better to thread data through the function hierarchy, so you don't need to worry about all the state lurking around.

app/src/main/java/com/example/bottomnav/DecodedData.java
```diff
@@ -0,0 +1,33 @@
...
+        }
+    }
+
+    static PayLoadDataType lookup(String name) {
+        for (PayLoadDataType dataType : PayLoadDataType.values()) {
+            if (dataType.name().equalsIgnoreCase(name)) {
+                return dataType;
+            }
+        }
```
ducky64: Can you inline this into the above (decodePrimitive)? And get rid of the intermediate PayloadDataType?

app/src/main/java/com/example/bottomnav/DecodedDouble.java
```diff
@@ -0,0 +1,9 @@
+package com.example.bottomnav;
+
+public class DecodedDouble implements DecodedData{
+    Double value;
+    @Override
+    public void decode() {
+
```
ducky64: Feel free to just ignore the double type, it's a data type that's pretty rare in embedded systems contexts.

app/src/main/java/com/example/bottomnav/DecodedFloat.java
```diff
@@ -0,0 +1,28 @@
...
+        int offset = 4;
+        int newSize = payload.length - offset;
+        byte[] packet = new byte[4];
+        bb.get(packet, 0, offset);
+        value = ByteBuffer.wrap(packet).order(ByteOrder.LITTLE_ENDIAN).getFloat();
+        if ((payload.length / offset) > 1) {
+            byte[] newPayload = new byte[newSize];
+            bb.get(newPayload, 0, newSize);
+            payload = newPayload;
+        }
```
ducky64: This seems to be doing a lot - I'm not sure how you're defining `decode` and what it's supposed to do.

I'd have written it to take in a byte buffer, and optionally output data (optional because the input may be malformed). In the simplest case, you could output a string (so the return type is consistent), if you wanted to be more fancy you could create a type hierarchy of decoded values.

Also note that each primitive data type has a fixed length. For example, `int` is 4 bytes here (but generally platform-dependent), `float` is 4 bytes, `double` is 8 bytes, and `uint8_t` is 1 byte.

I'm also not sure why this needs to modify the payload.

app/src/main/java/com/example/bottomnav/DecodedStruct.java
```diff
@@ -0,0 +1,27 @@
+package com.example.bottomnav;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+
+public class DecodedStruct implements DecodedData {
+    public HashMap<String, DecodedData> decodedValues = new HashMap<>();
+    ArrayList<StructContents> contents;
```
ducky64: One alternative is to have this delegate to DecodedData - so the contents is a list of DecodedData (with associated field name). Then, if DecodedData also provides a `getSize` method, the struct decoder can just iterate through the fields and advance the payload per field.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -79,28 +78,49 @@ private void storeStuct(IASTDeclaration[] declarations, String name) throws Exce
...
+
+    // Given a CAN ID name, retrieves its associated struct contents
+    public ArrayList<StructContents> getCanStruct(String canId) {
+        return getStructContents(canNameToStruct.get(canId));
+    }
+
+    // Give CAN ID, retrieves associated strict contents
+    public ArrayList<StructContents> getCanStruct(int canId) {
+        return getStructContents(canNameToStruct.get(canIdToName.get(canId)));
+    }
```
ducky64: I think you should only need the int version, since in actual decoding you'd only get the CAN ID as an integer. If you have a decoded name too, great, but you shouldn't need a name to decode the payload.

I also think in many cases, it's a good idea to build only the minimum interfaces you need for your current task, instead of trying implement everything that might be useful down the line (because it may turn out not to be useful, or may be obsoleted by restructuring)

app/src/main/java/com/example/bottomnav/Translation.java
```diff
@@ -0,0 +1,43 @@
+package com.example.bottomnav;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+
+public class Translation {
```
ducky64: I'd recommend for each class you write, add a short blurb that describes its purpose and what it does.

Writing summarizing comments is generally good programming practice, but also here it forces you to think - if it becomes difficult to write that comment (or if that comment becomes very verbose and confusing), it may suggest that refactoring is in order.

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -177,14 +177,47 @@ public void parseData() throws Exception {
...
     }
+
+    @Test
+    public void decodingSimple() throws Exception {
+
+        Parse parsedFile = Parse.parseTextFile("decode.h");
+        Translation test = new Translation(parsedFile);
+
+        byte[] packedFloatPayload = {0x71, (byte) 0xFD, 0x47, 0x41};
+        test.decode(0x310, packedFloatPayload);
```
ducky64: Does this return something, that you want to check the value of?

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -177,14 +177,47 @@ public void parseData() throws Exception {
...
+        assertEquals(12.499375343322754, float1.value, 0.000000000000001);
+
+        byte[] var4Payload = {0x0B, 0x00, 0x00, (byte) 0xA5};
+        DecodedStruct struct = (DecodedStruct) test.decode(0x282, var4Payload);
+        DecodedInteger int1 = (DecodedInteger) struct.getValue("accelPos");
+        DecodedInteger int2 = (DecodedInteger) struct.getValue("brakePos");
+        DecodedInteger int3 = (DecodedInteger) struct.getValue("reserved1Pos");
+        DecodedInteger int4 = (DecodedInteger) struct.getValue("reserved2Pos");
+
+        assertEquals(11, int1.value);
```
ducky64: Can you inline this? so it's more visually obvious what the expected value goes with?
`assertEquals(11, (DecodedInteger) struct.getValue("accelPos"));`

ducky64 (commented):
Good progress, though I think there's more that can be done to clean up the code.
- In general, less code is better (within reason - where it enhances readability instead of becoming code golf) - see where it makes sense to inline functions and when intermediate data types and storage isn't necessary.
- In general, I'd recommend summarizing comments for each class. See inline review comment for more details.
- When I suggested the type hierarchy structure, I was envisioning something more like `DataDecoder` - the can_id.h parser creates a DataDecoder object for each `@canPayloadStruct` or `@payloadDataType` annotation, and that is stored in a can ID -> DataDecoder HashMap. Then, when we receive a CAN packet, we can get the relevant DataDecoder and call its `.decode(...)`, passing in the payload. This can handle both the struct and primitive data type case in a unified structure.
  - There would be the question of what type `.decode(...)` returns. In the simplest case it can just be a string (good enough for now, where we just want a human-readable representation), but in the future we can think of richer data types, for example an actual `DecodedData` class that has types like `NumericData` which (for example) can be graphed.
- Using global variables to pass around transient data is a huge red flag, since it's hard to reason about what's going on where.

Let me know if you'd like to discuss in more detail at some point.

app/src/main/java/com/example/bottomnav/DataDecoder.java
```diff
@@ -0,0 +1,56 @@
+package com.example.bottomnav;
+
+public interface DataDecoder {
+    static DataDecoder getPrimativeDecoder(PayLoadDataType dataType) {
```
ducky64: I'd prefer this returned an Optional<DataDecoder> to explicitly indicate that it can fail

app/src/main/java/com/example/bottomnav/DoubleDecoder.java
```diff
@@ -0,0 +1,16 @@
+package com.example.bottomnav;
+
+public class DoubleDecoder implements DataDecoder {
+    public int packetSize = 2;
+    public double value;
+
+    @Override
+    public String decode(Integer canId, byte[] payload) {
+        return null;
```
ducky64: TODO implement me?

app/src/main/java/com/example/bottomnav/DataDecoder.java
```diff
@@ -0,0 +1,56 @@
...
+            case uint8_t:
+                return new IntegerDecoder();
+            case uint16_t:
+                return new IntegerDecoder();
+            default:
+                return null;
+        }
+    }
+
+    static String getPrimativeValue(PayLoadDataType dataType, DataDecoder decoder) {
```
ducky64: I'd recommend refactoring this into each decoder's `decodeToString` method or something.

In general, I'd say that if you're trying to define specialized behavior for objects on a class hierarchy, define it on the classes by overriding. In `getPrimitiveDecoder` you don't have this option, since you don't have the decoder yet, but here you have the decoder.

app/src/main/java/com/example/bottomnav/IntegerDecoder.java
```diff
@@ -0,0 +1,22 @@
+package com.example.bottomnav;
+
+import java.nio.ByteBuffer;
+
+public class IntegerDecoder implements DataDecoder {
```
ducky64: Because you're going to have a few data types that look like integer (eg, uint8_t, uint16_t, ...), maybe this could be a class that's parameterized by the byte width of the data type, to avoid a copypasting IntegerDecoder, LongDecoder, and whatnot?

app/src/main/java/com/example/bottomnav/PayLoadDataType.java
ducky64: Can you get rid of this? This seems like an intermediary between the raw string type and the DataDecoder. Unless you have a compelling reason for needing an intermediate data type, I'd just suggest choosing one of those two to use elsewhere?

In general, you want to minimize the amount of code - every line is a maintenance burden, and additional data types are another thing that needs to be learned to work with the codebase.

app/src/main/java/com/example/bottomnav/StructDecoder.java
```diff
@@ -0,0 +1,59 @@
+package com.example.bottomnav;
+
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.HashMap;
+
+public class StructDecoder implements DataDecoder {
+    private ArrayList<StructContents> contents;
+    public  HashMap<String, DataDecoder> decodedPrimatives = new HashMap<>();
+    private StringBuilder message = new StringBuilder();
```
ducky64: Does this actually need to be persistent? I think it would actually be bad for this to be persistent, right?

app/src/main/java/com/example/bottomnav/StructDecoder.java
```diff
@@ -0,0 +1,59 @@
+package com.example.bottomnav;
+
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.HashMap;
+
+public class StructDecoder implements DataDecoder {
+    private ArrayList<StructContents> contents;
+    public  HashMap<String, DataDecoder> decodedPrimatives = new HashMap<>();
```
ducky64: This should be private or protected, this should be an internal detail

app/src/main/java/com/example/bottomnav/StructDecoder.java
```diff
@@ -0,0 +1,59 @@
+package com.example.bottomnav;
+
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.HashMap;
+
+public class StructDecoder implements DataDecoder {
+    private ArrayList<StructContents> contents;
```
ducky64: Would it make sense to merge StructContents into here? Because StructContents doesn't really have a reason to exist independently from this right?

Similarly with PayloadDataType, StructContents is kind of an intermediate format, and (if the code can actual work out) I'd prefer to encapsulate the decoding logic in here. We can discuss details about how to implement this.

app/src/main/java/com/example/bottomnav/StructDecoder.java
```diff
@@ -0,0 +1,59 @@
...
+            PayLoadDataType type = DataDecoder.lookup(variable.type);
+            DataDecoder primativeDecoder = DataDecoder.getPrimativeDecoder(type);
+            primativeDecoder.decode(canId, payload);
+            decodedPrimatives.put(variable.name, primativeDecoder);
+            payload = adjustPayload(payload, primativeDecoder.getPacketSize());
+        }
+        return getDecodedValues();
+    }
+
+    public byte[] adjustPayload(byte[] payload, int packetSize) {
```
ducky64, line 48: The logic here seems a bit complicated, I'm not sure what you're trying to do here.
Remember that if the input is invalid, you should fail in a way that isn't catastrophic (crashing the app - because we will probably get scrambled data once in a while), but also shouldn't create misleading data out of nowhere.
We can discuss how to do error handling.

Also, is [Array.copyOfRange](https://stackoverflow.com/a/11001759/5875811) what you're looking for?

app/src/main/java/com/example/bottomnav/StructDecoder.java
```diff
@@ -0,0 +1,59 @@
...
+        return message.toString();
+    }
+
+    public DataDecoder getValue(String variableName) {
+        return decodedPrimatives.get(variableName);
+    }
+
+    @Override
+    public int getPacketSize() {
+        return 0;
```
ducky64: Make sure to actually implement these, or make sure they fail noisily (for example throwing a not implemented exception or something) - you don't want to leave submarine TODOs through your code.

The suggestion here would just be to sum the contents DataDecoder's getPacketSize,

ducky64 (commented):
Better! But I think you can still move more code into the DataDecoder class, where it will make more sense.

See rest of the inline comments, we can also discuss in more detail.

app/src/main/java/com/example/bottomnav/Components.java
```diff
@@ -5,39 +5,107 @@
...
+    /**
+     *
+     * @param simpleSpec
+     * @param declarator
+     * @return Optional<Components>
+     * @throws Exception
+     *
+     * Difference: Simple specifiers don't quantifier types and do not have name object
+     */
+    private static Optional<Components> makeOptComp(CPPASTSimpleDeclSpecifier simpleSpec,
```
ducky64: I'd recommend naming this something more descriptive, perhaps makeComponentFromDecl, or alternatively just `fromDecl`, since it's a static method so the Component part can be implicit

app/src/main/java/com/example/bottomnav/Components.java
```diff
@@ -5,39 +5,107 @@
...
+        } else {
+            return Optional.empty();
+        }
+    }
+
+    /**
+     * This parser is a little goofy, instead of returning toString version of prim. type, it
+     * returns a number
+     */
+    private static char[] getPrimitiveType(int primType) {
```
ducky64: This is weird - does the stock AST parser actually map types to a number? If so - please make it clear in the comments that this is the workaround for the AST parser, instead of this parser - which implies your code (that we can fix) instead of external library code (that would be more of an uphill battle to fix)

app/src/main/java/com/example/bottomnav/Components.java
```diff
@@ -5,39 +5,107 @@
...
+     *
+     * Difference: Name typed specifiers have quantifer types and have name object
+     */
+    private static Optional<Components> makeOptComp(CPPASTNamedTypeSpecifier namedTypeSpec,
+                                                    CPPASTDeclarator declarator) throws Exception {
         CPPASTEqualsInitializer init = (CPPASTEqualsInitializer) declarator.getInitializer();
-        Optional<CPPASTEqualsInitializer> initOpt = makeOpt(init);
-        if (declarator != null && specifier != null && declaration != null) {
+        Optional<CPPASTEqualsInitializer> initOpt = makeOptInit(init);
+        CPPASTName TQname = new CPPASTName();
```
ducky64: This is a bit weird that you have to materialize a new AST node. Can you refactor your interfaces so you don't have to do this? For example, instead of defining Component on AST nodes, define it using the unpacked data like strings?

app/src/main/java/com/example/bottomnav/FloatDecoder.java
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    @Override
+    public String decode(Integer canId, byte[] payload) {
+        ByteBuffer bb = ByteBuffer.wrap(payload);
+        byte[] packet = new byte[packetSize];
+        bb.get(packet, 0, packetSize);
+        rawValue = (T) new Float(ByteBuffer.wrap(packet).order(ByteOrder.LITTLE_ENDIAN).getFloat());
+        value = "" + payloadDataType + ": " + rawValue;
+        return value;
```
ducky64: Is there a reason this can't look exactly like the DoubleDecoder version, except w/ getFloat?

app/src/main/java/com/example/bottomnav/DoubleDecoder.java
```diff
@@ -0,0 +1,17 @@
+package com.example.bottomnav;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+public class DoubleDecoder<T> extends IntegerDecoder {
```
ducky64: I don't think this should extend IntegerDecoder. Extending (subclassing) means an is-a relation, which this is not. This uses some shared methods, but it fundamental is not an integer decoder. You might consider having an intermediate class in the hierarchy, eg PrimitiveDecoder, that is defined as taking the next n contiguous bytes, maybe decoding them as some format, then some some abstract transformation. Both Float, Double, and IntegerDecoders would implement that type.

For example:
```java
class PrimitveDecoder extends DataDecoder {
  public PrimitiveDecoder(int byteSize, String dataType) {
    ...
  }
  String decode(...) {
    ByteBuffer bytes = ...;  // this just gets the next n bytes
    transform(bytes)  // then leaves it to the subclass to do something with them
  }
  
  public T transform(ByteBuffer bytes)  // to be implemented by subclass
}
```

app/src/main/java/com/example/bottomnav/IntegerDecoder.java
```diff
@@ -0,0 +1,41 @@
...
+        payloadDataType = dataType;
+        packetSize = byteSize;
+    }
+
+    @Override
+    public String decode(Integer canId, byte[] payload) {
+        ByteBuffer bb = ByteBuffer.wrap(payload);
+        byte[] packet = new byte[packetSize];
+        bb.get(packet, 0, packetSize);
+        rawValue = (T) new Integer(new BigInteger(packet).intValue() & 0xff);
```
ducky64: Using mutable state is highly not recommended, because now calling valueToString requires calling decode right before - a requirement that isn't obvious or intuitive.

I recommend delegating instead:
```java
public String decode(...) {  // maybe rename decodeToString or something?
  return payloadDataType + ": " + valueToRaw(canId, payload)
}

public T valueToRaw(...) {  // would need to take canId and payload
  // actual decoding happens here
}
```

Also, is this somewhat duplicated w/ valueToString? Consider getting rid of one of them - and leaving it to the next level (caller) if it wants to do some specialty formatting. Try to keep your classes as single responsibility as possible.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -19,12 +19,12 @@
 import java.util.regex.Pattern;
 
 public class Parse {
-    // Const name -> ConstContents
-    public HashMap<String, ConstContents> constRepo = new HashMap<>();
-    // Struct name -> StructContents
-    public HashMap<String, ArrayList<StructContents>> structRepo = new HashMap<>();
-    // CAN ID -> CAN Struct
-    public HashMap<String, String> IDToStruct = new HashMap<>();
+    private HashMap<String, ConstContents> constRepo = new HashMap<>();
```
ducky64: This is a lot of stuff to store, at some point we should revisit this to determine if we can reduce this down

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -50,57 +50,88 @@ public Parse(char[] code) throws Exception {
...
+        } else {
+            return "single " + canIdToName.get(canID) + ", "
+                    + payloadMap.get(canID).get().decode(canID, payload);
         }
     }
 
-    // Given a CAN ID name, retrieves its associated struct contents
-    public ArrayList<StructContents> getCanStruct(String idName) {
-        return getStruct(IDToStruct.get(idName));
+    public DataDecoder getDecoder(Integer canID) {
```
ducky64: I'd recommend making this return Optional to make it clear it can fail

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -50,57 +50,88 @@ public Parse(char[] code) throws Exception {
...
 
-    private void associateIDStruct(IASTNode[] comments) {
-        Pattern pattern = Pattern.compile("@canPayloadStruct\\s+(\\S+)\\s*=\\s*(\\S+)");
-        for (IASTNode comment : comments) {
-            Matcher matcher = pattern.matcher(comment.getRawSignature());
-            if (matcher.find()) {
-                IDToStruct.put(matcher.group(1), matcher.group(2));
-            }
+    public String decode(Integer canID, byte[] payload) {
+        if (canIDToStruct.containsKey(canID)) {
```
ducky64: What happens if there is no decoding definition?

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -109,82 +103,130 @@ public void structNConst() throws Exception {
...
+        IntegerDecoder floatDec = (IntegerDecoder) test.getDecoder(0x310);
+        assertEquals(12.499375343322754,(float) floatDec.valueToRaw(), 0.000000000000001);
+
+        byte[] var4Payload = {0x0B, 0x00, 0x00, (byte) 0xA5};
+        String var4PaylodMessage = "struct CanPedalPosStruct, \n" +
+                "    accelPos, uint8_t: 11\n" +
+                "    brakePos, uint8_t: 0\n" +
+                "    reserved1Pos, uint8_t: 0\n" +
+                "    reserved2Pos, uint8_t: 165";
+        System.out.println(test.decode(0x282, var4Payload));
```
ducky64: You generally don't want printlns in your test code. It's fine for debugging, but unit tests are meant to be run fully automated aside from the pass/fail at the end. tl;dr remove stray debugging printlns in final code

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -109,82 +103,130 @@ public void structNConst() throws Exception {
...
+
+    @Test
+    public void checkSimple() {
+        byte[] var4Payload = {0x0B, 0x00, 0x00, (byte) 0xA5};
+        byte[] ok = {0x0B};
+
+
+        int val = new BigInteger(ok).intValue();
+        System.out.println(val);
+        //ByteBuffer.wrap(var4Payload).order(ByteOrder.LITTLE_ENDIAN).
```
ducky64: Similarly, remove dead code

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -23,23 +23,18 @@ public void simpleCheck() throws Exception {
...
-        assertEquals("uint16_t", test.getConst("CAN_ID").type);
-        assertEquals("uint16_t", test.getConst("CAN_ID2").type);
-        assertEquals("uint16_t", test.getConst("CAN_ID3").type);
-
-        assertEquals("0x16", test.getConst("CAN_ID").value);
-        assertEquals("0x32", test.getConst("CAN_ID2").value);
-        assertEquals("0x64", test.getConst("CAN_ID3").value);
+        assertEquals("0x16", test.getConstContents("CAN_ID").value);
+        assertEquals("0x32", test.getConstContents("CAN_ID2").value);
+        assertEquals("0x64", test.getConstContents("CAN_ID3").value);
```
ducky64: Can you also add a test somewhere for if we try to decode a CAN ID that doesn't exist? We do not have all the CAN IDs defined in can_struct.h, and in any case the parser should be robust to when data is unavailable.

ducky64 (commented):
Overall looks better. Main suggestions (see inline comments):
- Make sure when you subclass, you follow the is-a relation. You may find it useful to introduce an intermediate class in the class hierarchy
- Avoid state where possible - in IntegerDecoder, get rid of rawValue and replace its use with a eg valueToRaw call. State can be confusing to work around.
- Make sure the decoder is robust to when there is no struct / ID definition available.

ducky64 (approved):
I think this is looking reasonable, especially since we're at the track now and need to integrate.

If you're interested after the race, I'd like to get this cleaned up further - there's a few more items that could be architecturally cleaner and more robust.

PR #8, opened 2022-06-16
Can data finished

BAKA-TRASHPANDA:
Checking CAN DATA for merge to main

### DIFF
--- a/.idea/gradle.xml
+++ b/.idea/gradle.xml
@@ -7,6 +7,7 @@
         <option name="testRunner" value="GRADLE" />
         <option name="distributionType" value="DEFAULT_WRAPPED" />
         <option name="externalProjectPath" value="$PROJECT_DIR$" />
+        <option name="gradleJvm" value="16" />
         <option name="modules">
           <set>
             <option value="$PROJECT_DIR$" />
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -15,7 +15,7 @@
       </map>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_11" default="true" project-jdk-name="Android Studio default JDK" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="true" project-jdk-name="Android Studio default JDK" project-jdk-type="JavaSDK">
     <output url="file://$PROJECT_DIR$/build/classes" />
   </component>
   <component name="ProjectType">
--- a/app/src/main/java/com/example/bottomnav/CAN_Data.java
+++ b/app/src/main/java/com/example/bottomnav/CAN_Data.java
@@ -1,0 +1,65 @@
+package com.example.bottomnav.ui.notifications;
+
+class CAN_Data {
+
+    Integer id;
+    Integer data;
+    Integer len;
+    public CAN_Data(Integer id,Integer data, Integer len ){
+        this.id = id;
+        this.data = data;
+        this.len = len;
+    }
+    public static CAN_Data decode(String raw){
+        Integer id = -1;
+        Integer len;
+        Integer data;
+        int index = -1;
+        if(raw.substring(0,1).equals('t') || raw.substring(0,1).equals('r')){
+            id = Integer.parseInt(raw.substring(1, 4), 16);
+            index = 4;
+        }
+        else if(raw.substring(0,1).equals('T') || raw.substring(0,1).equals('R')){
+            id = Integer.parseInt(raw.substring(1, 9), 16);
+            index = 9;
+        }
+        else{
+            return null;
+        }
+
+        if((index != -1 ) && (id != -1)) {
+            len = Integer.parseInt(raw.substring(index, index + 1), 16);
+            data = Integer.parseInt(raw.substring(index + 1, raw.length()), 16);
+            CAN_Data decoded = new CAN_Data(id, data, len);
+            return decoded;
+        }
+        else{
+            return null;
+        }
+
+    }
+    public Integer getId(){
+        return id;
+    }
+
+    public Integer getData() {
+        return data;
+    }
+
+    public Integer getLen() {
+        return len;
+    }
+
+    public void setData(Integer data) {
+        this.data = data;
+    }
+
+    public void setId(Integer id) {
+        this.id = id;
+    }
+
+    public void setLen(Integer len) {
+        this.len = len;
+    }
+
+}
--- a/app/src/main/java/com/example/bottomnav/CAN_Data.java
+++ b/app/src/main/java/com/example/bottomnav/CAN_Data.java
@@ -2,19 +2,17 @@ package com.example.bottomnav;
 
 class CAN_Data {
     int id;
-    int data;
-    int len;
+    int[] data;
 
-    public CAN_Data(int id, int data, int len) {
+    public CAN_Data(int id, int[] data, int len) {
         this.id = id;
         this.data = data;
-        this.len = len;
     }
 
     public static CAN_Data decode(String raw) {
         int id = -1;
         int len;
-        int data;
+
         int index = -1;
         if (raw.charAt(0) == 't' || raw.charAt(0) == 'r') {
             id = Integer.parseInt(raw.substring(1, 4), 16);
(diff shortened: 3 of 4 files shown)


.idea/gradle.xml
```diff
@@ -7,6 +7,7 @@
         <option name="testRunner" value="GRADLE" />
         <option name="distributionType" value="DEFAULT_WRAPPED" />
         <option name="externalProjectPath" value="$PROJECT_DIR$" />
+        <option name="gradleJvm" value="16" />
```
ducky64, line 10: Can you back out this unrelated change?

.idea/misc.xml
```diff
@@ -15,7 +15,7 @@
       </map>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_11" default="true" project-jdk-name="Android Studio default JDK" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="true" project-jdk-name="Android Studio default JDK" project-jdk-type="JavaSDK">
```
ducky64, line 18: Ditto, can you back this out?

app/src/test/java/com/example/bottomnav/DataTest.java
```diff
@@ -0,0 +1,24 @@
...
+    }
+    @Test
+    public void BIGData() {
+        String data = "t0568111111111111111A";
+        CAN_Data newData = CAN_Data.decode(data);
+        System.out.println(newData.getData());
+        assertEquals( 0x11, newData.getData()[0]);
+        assertEquals( 8, newData.getData().length);
+    }
+}
```
ducky64, line 24: To be thorough, can you also add tests for the extended case?

In general, I'd advocate unit testing all code paths that are easily testable. I believe that in software engineering practice it isn't uncommon to have as much (or even more) testing code as actual application code.

Also, consider testing malformed inputs. Because of the nature of a wireless connection, it may be possible to get partial (possibly even corrupted) data, so your parser needs to be tolerant. It doesn't need to decode the data successfully (which would be impossible if it were corrupted), but it does need to not crash and not give misleading results.

app/src/test/java/com/example/bottomnav/DataTest.java
```diff
@@ -0,0 +1,24 @@
...
+        System.out.println(newData.getData());
+        assertEquals( 0x11 , newData.getData()[0]);
+        assertEquals( 86 , newData.getId());
+    }
+    @Test
+    public void BIGData() {
+        String data = "t0568111111111111111A";
+        CAN_Data newData = CAN_Data.decode(data);
+        System.out.println(newData.getData());
+        assertEquals( 0x11, newData.getData()[0]);
```
ducky64, line 21: I'd recommend checking all the payload data.
And all the parsed data - including ID.

app/src/test/java/com/example/bottomnav/DataTest.java
```diff
@@ -0,0 +1,24 @@
...
+        CAN_Data newData = CAN_Data.decode(data);
+        System.out.println(newData.getData());
+        assertEquals( 0x11 , newData.getData()[0]);
+        assertEquals( 86 , newData.getId());
+    }
+    @Test
+    public void BIGData() {
+        String data = "t0568111111111111111A";
+        CAN_Data newData = CAN_Data.decode(data);
+        System.out.println(newData.getData());
```
ducky64, line 20: Unit tests shouldn't have printf, unit tests should be fully automated

app/src/main/java/com/example/bottomnav/CAN_Data.java
```diff
@@ -0,0 +1,48 @@
...
+        if (raw.charAt(0) == 't' || raw.charAt(0) == 'r') {
+            id = Integer.parseInt(raw.substring(1, 4), 16);
+            index = 4;
+        }
+        else if (raw.charAt(0) == 'T' || raw.charAt(0) == 'R') {
+            id = Integer.parseInt(raw.substring(1, 9), 16);
+            index = 9;
+        }
+        else {
+            return null;
```
ducky64, line 26: `nulls` can lead debugging nightmares, because there's nothing in the method signature that indicates that it can be null, so more often then not one might just assume that it'll be non-null.

Where you have a code path that can fail, I recommend encoding the possibility of failure in the type system, such as through an [Optional type](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html). I believe this is is more common in functional languages, but is more supported elsewhere now (including being part of the Java language)

app/src/main/java/com/example/bottomnav/CAN_Data.java
```diff
@@ -0,0 +1,48 @@
...
+    public CAN_Data(int id, int[] data, int len) {
+        this.id = id;
+        this.data = data;
+    }
+
+    public static CAN_Data decode(String raw) {
+        int id = -1;
+        int len;
+
+        int index = -1;
```
ducky64, line 16: Can you leave this uninitialized? It doesn't really need to be -1?
Similarly with `id`.

Hopefully if the IDE is good, it can also warn you if you end up using an uninitialized variable, whereas if you give it a default it won't be able to warn you.

app/src/main/java/com/example/bottomnav/CAN_Data.java
```diff
@@ -0,0 +1,48 @@
...
+        }
+        else if (raw.charAt(0) == 'T' || raw.charAt(0) == 'R') {
+            id = Integer.parseInt(raw.substring(1, 9), 16);
+            index = 9;
+        }
+        else {
+            return null;
+        }
+
+        len = Integer.parseInt(raw.substring(index, index + 1), 16);
```
ducky64, line 29: Consider adding a validation step to make sure:
- that the input is long enough such that it has a length field
- that the length is 8 or less
- that the input is long enough for the entire payload

and if it fails validation, return `Optional.empty()`

Also, here (and every other instance of `parseInt`) consider having error handling, specifically catching the `NumberFormatException` that `parseInt` can throw and returning Optional.empty(). You might as well wrap the entire method in a try-catch or something, since you use a lot of parseInt.

app/src/main/java/com/example/bottomnav/CAN_Data.java
```diff
@@ -0,0 +1,48 @@
...
+        this.data = data;
+    }
+
+    public static CAN_Data decode(String raw) {
+        int id = -1;
+        int len;
+
+        int index = -1;
+        if (raw.charAt(0) == 't' || raw.charAt(0) == 'r') {
+            id = Integer.parseInt(raw.substring(1, 4), 16);
```
ducky64, line 18: Consider adding validation here, to check that the input contains enough characters to decode the ID, and optionally the resulting decoded ID is within the 11-bit standard ID format defined by CAN.

app/src/main/java/com/example/bottomnav/CAN_Data.java
```diff
@@ -0,0 +1,48 @@
...
+        }
+        else {
+            return null;
+        }
+
+        len = Integer.parseInt(raw.substring(index, index + 1), 16);
+        int[] data = new int[len];
+
+        for(int x = 0;x<len;x+=1) {
+            data[x] = Integer.parseInt(raw.substring(2*x +index+1, 2*x+2+index+1), 16);
```
ducky64, line 33:
```suggestion
            data[x] = Integer.parseInt(raw.substring(ndex+1 + 2*x, index+1 + 2*x + 2), 16);
```
I think this should clarify the indexing a bit more?

app/src/main/java/com/example/bottomnav/CAN_Data.java
```diff
@@ -0,0 +1,48 @@
...
+    public int getId(){
+        return id;
+    }
+
+    public int[] getData() {
+        return data;
+    }
+
+
+
```
ducky64, line 47:
```suggestion
```

excessive newlines can be deleted

ducky64 (changes requested):
Overall, this looks like it will work fine, but I think it could use some robustness (see inline comments). Especially since data is coming in from wireless, it may be possible to get malformed or corrupted data, and your parser should be tolerant of this (ie not crash), even if all it does is throw up its hands and return an error or empty type.

PR #9, opened 2022-08-31
Decode Clean

MichaelCampos-eng:
The decoders are now stateless. Each decoder has a decode to raw funciton and a decode to string function. Decode to string returns an optional of a string, which contains the name of the variable and its value. A static function was added to parse the output string to retrieve the name and value.


app/src/main/java/com/example/bottomnav/DataDecoder.java
```diff
@@ -1,52 +1,85 @@
...
+import java.util.regex.Pattern;
 
 public interface DataDecoder<T> {
 
     static Optional<DataDecoder> createPrimitiveDecoder(VariableContents contents) {
         switch (contents.payloadDataType) {
             case "int":
-                contents.setPacketSize(1);
-                return Optional.of(new IntegerDecoder<Integer>(contents));
+                return Optional.of(new IntegerDecoder<Integer>(1, contents.name));
```
ducky64: You may want to double check this but I think int defaults to 32 bits (4 bytes).

Note that the C standard doesn't define the exact size of int, so specific behavior is per-compiler.

app/src/main/java/com/example/bottomnav/DataDecoder.java
```diff
@@ -1,52 +1,85 @@
...
-                contents.setPacketSize(8);
-                return Optional.of(new DoubleDecoder<Double>(contents));
+                return Optional.of(new DoubleDecoder<Double>(8, contents.name));
             default:
                 return Optional.empty();
         }
     }
 
-    String decode(Integer canId, byte[] payload);
+    static Optional<DataDecoder> createStructDecoder(ArrayList<VariableContents> variables) {
```
ducky64: Since this can only create a StructDecoder, perhaps this should live inside StructDecoder?

app/src/main/java/com/example/bottomnav/DataDecoder.java
```diff
@@ -1,52 +1,85 @@
...
+        public String getName() {
+            return nameVariable;
+        }
+
+        public String getValue() {
+            return decodedValue;
+        }
+    }
+
+    static ArrayList<Decoded> parseStringResult(String resultOfDecoder) {
```
ducky64: A comment here with input-output examples and applications would be super helpful

app/src/main/java/com/example/bottomnav/IntegerDecoder.java
```diff
@@ -1,20 +1,16 @@
...
     @Override
-    public String decode(Integer canId, byte[] payload) {
-        ByteBuffer bb = ByteBuffer.wrap(payload);
-        byte[] packet = new byte[contents.packetSize];
-        bb.get(packet, 0, contents.packetSize);
-        rawValue = (T) new Integer(new BigInteger(packet).intValue());
-        value = "" + rawValue;
-        return valueToString();
+    public Optional<T> getRawValue(byte[] payload) {
+        return Optional.of((T) new Integer(new BigInteger(wrapPayload(payload)).intValue()));
```
ducky64: Doesn't size need to be taken into account here somewhere?
(what if the input is too long?)

app/src/main/java/com/example/bottomnav/MainActivity.java
```diff
@@ -1,17 +1,32 @@
...
 import androidx.appcompat.app.AppCompatActivity;
+import androidx.core.app.ActivityCompat;
 import androidx.navigation.NavController;
 import androidx.navigation.Navigation;
 import androidx.navigation.ui.AppBarConfiguration;
 import androidx.navigation.ui.NavigationUI;
+
+import com.example.bottomnav.bluetoothlegatt.DeviceControlActivity;
+import com.example.bottomnav.databinding.ActivityMainBinding;
+import com.example.bottomnav.BluetoothLeService;
```
ducky64: This seems orthogonal to the main PR feature - not great practice but fine here.

That being said, are all these imports necessary? In IntelliJ, if these are greyed out, can you hit "optimize imports" to clean these up?

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -51,75 +53,57 @@ public Parse(char[] code) throws Exception {
...
-        if (data.isPresent() && data.get().init.isPresent()) {
-            Components comp = data.get();
-            CPPASTLiteralExpression value =
-                    (CPPASTLiteralExpression) comp.init.get().getInitializerClause();
-            VariableContents contents = new VariableContents(comp.name, value.getRawSignature(),
-                    comp.typeQualifier, comp.type);
-            Optional<DataDecoder> decoder = DataDecoder.createPrimitiveDecoder(contents);
-            decoderRepo.put(contents.name, decoder);
-            canIdToName.put(Integer.decode(contents.value), contents.name);
+        Optional<VariableContents> data = VariableContents.getContents(declaration);
```
ducky64: A short comment (associated with this function) would also be useful to summarize what this does and the expected preconditions (eg, should only be used with a `const` declaration.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -51,75 +53,57 @@ public Parse(char[] code) throws Exception {
...
+        if (data.isPresent() && data.get().value.isPresent()) {
+            VariableContents variableContents = data.get();
+            decoderRepo.put(variableContents.name, DataDecoder.createPrimitiveDecoder(variableContents));
```
ducky64: What is DataDecoder.createPrimitiveDecoder returns Optional.empty?

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -51,75 +53,57 @@ public Parse(char[] code) throws Exception {
...
 
     public Optional<DataDecoder> getDecoder(String name) {
-        if (canNameToStruct.containsKey(name)) {
-            return decoderRepo.get(canNameToStruct.get(name));
-        } else if (decoderRepo.containsKey(name)) {
-            return decoderRepo.get(name);
-        } return Optional.empty();
+        if (!decoderRepo.containsKey(name)) {
+            return Optional.empty();
+        } return decoderRepo.get(name);
```
ducky64:
```suggestion
        }
        return decoderRepo.get(name);
```
style

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -51,75 +53,57 @@ public Parse(char[] code) throws Exception {
...
+        return getDecoder(IdToName.get(canId));
     }
 
-    public VariableContents getConstContents(String name) {
-        return ((PrimitiveDecoder) decoderRepo.get(name).get()).contents;
-    }
-
-    public ArrayList<VariableContents> getStructContents(String name) {
-        return ((StructDecoder) getDecoder(name).get()).variables;
+    public Optional<String> decodeToString (Integer canId, byte[] payload) {
```
ducky64: I think this is kind of redundant with getDecoder?

app/src/main/java/com/example/bottomnav/PrimitiveDecoder.java
```diff
@@ -1,54 +1,43 @@
...
-    @Override
-    public int getSize() {
-        return size;
-    }
-
-    @Override
-    public String valueToString() {
-        return "" + contents.name + ": " + value;
-    }
+public class PrimitiveDecoder<T> implements DataDecoder {
```
ducky64: A comment here could help too, describing what this does and example usages, especially since this is an interface meant to be extended (inherited)

app/src/main/java/com/example/bottomnav/PrimitiveDecoder.java
```diff
@@ -1,54 +1,43 @@
...
+    protected int packetSize;
```
ducky64: Naming: I don't think packet is the right term, since a primitive decoder may be part of a struct that forms the complete packet payload. Maybe a more precise name is `variableSize` or `typeSize`?

app/src/main/java/com/example/bottomnav/PrimitiveDecoder.java
```diff
@@ -1,54 +1,43 @@
...
+        if (rawOptional.isPresent()) {
+            return Optional.of(variableName + ": " + rawOptional.get());
+        } else {
+            return Optional.empty();
+        }
     }
 
-    public String getValueString() {
-        return value;
+    public Optional<T> getRawValue(byte[] payload) {
```
ducky64: Instead of having this return null can you just make this method abstract?

That way the type checker can provide errors if subclasses fail to override this, which is probably an error.

Additionally, would it be more grammatically parallel (with decodeToRaw) and precise to name this `decodeRawPrimitive` or something?

app/src/main/java/com/example/bottomnav/PrimitiveDecoder.java
```diff
@@ -1,54 +1,43 @@
...
+        return null;
     }
 
-    public int getPacketSize() {
-        return contents.packetSize;
+    public byte[] wrapPayload(byte[] payload) {
```
ducky64: Can you add a comment on the behavior of this and what this is for?

I'm also not convinced it's a good idea to trim and reformat data, in part because it may give the user misleading information about data being cleaner and more well formatted than it actually might be on the wire. We already have the infrastructure to fail noisily (but not catastrophically) in the case of bad input using Optional.empty.

app/src/main/java/com/example/bottomnav/StructDecoder.java
```diff
@@ -6,42 +6,47 @@
 import java.util.Optional;
 
 public class StructDecoder implements DataDecoder {
-    private HashMap<String, DataDecoder> decodedPrimatives = new HashMap<>();
-    public ArrayList<VariableContents> variables;
-    private int size;
+    private ArrayList<Optional<DataDecoder>> decodedPrimatives;
```
ducky64:
```suggestion
    private ArrayList<DataDecoder> decodedPrimitives;
```

I don't think this should be optional - there should not be a case where this is created with an unknown field. Partially because data sanitization should be handled at a higher level (before it becomes an internal variable) and partly because if there is an unknown field of unknown size that messes up any subsequent decoding.

Having your data sanitized before this also means you can clean up any downstream logic by removing the Optional tests.

Also, spelling.

app/src/main/java/com/example/bottomnav/StructDecoder.java
```diff
@@ -6,42 +6,47 @@
...
-        }
-        return message.toString();
+    public StructDecoder(ArrayList<Optional<DataDecoder>> decoded) {
+        decodedPrimatives = decoded;
     }
 
     @Override
-    public Object valueToRaw() {
-        return null;
+    public Optional<ArrayList> decodeToRaw(byte[] payload) {
```
ducky64:
```suggestion
    public Optional<ArrayList<Optional>> decodeToRaw(byte[] payload) {
```

While you don't known the type parameter of the Optional you still do know that it is Optionals

app/src/main/java/com/example/bottomnav/StructDecoder.java
```diff
@@ -6,42 +6,47 @@
...
+                Optional<String> decoded = decoder.get().decodeToString(payload);
+                if (decoded.isPresent()) {
+                    resultString += decoded.get() + tail;
+
+                } else {
+                    return Optional.empty();
+                }
+                payload = adjustPayload(payload, ((PrimitiveDecoder) decoder.get()).packetSize);
+            } else {
+                return Optional.empty();
```
ducky64: Consider being able to return a partial decode, instead of failing completely if any element fails to decode.

app/src/main/java/com/example/bottomnav/StructDecoder.java
```diff
@@ -6,42 +6,47 @@
...
+        return Optional.of(rawOptionals);
     }
 
     @Override
-    public String decode(Integer canId, byte[] payload) {
-        for (VariableContents variable : variables) {
-            Optional<DataDecoder> decoder = DataDecoder.createPrimitiveDecoder(variable);
+    public Optional<String> decodeToString(byte[] payload) {
+        String resultString = "";
+        String tail = ", ";
```
ducky64: Consider using [String.join](https://stackoverflow.com/a/22577565/5875811)!
You will have to convert all the decoded primitives to an ArrayList<String> first, but that simplifies a lot of this [fenceposting](https://en.wikipedia.org/wiki/Off-by-one_error#Fencepost_error) logic.

app/src/main/java/com/example/bottomnav/UnsignedIntegerDecoder.java
```diff
@@ -1,25 +1,18 @@
...
-        rawValue = (T) new Integer(new BigInteger(packet).intValue() & sign);
-        value = "" + rawValue;
-        return valueToString();
+    public Optional<T> getRawValue(byte[] payload) {
+        return Optional.of((T) new Integer(new BigInteger(wrapPayload(payload)).intValue() & sign));
     }
-
-
-}
+}
```
ducky64:
```suggestion
}

```

recommended style convention: newlines at end of file

app/src/main/java/com/example/bottomnav/UnsignedIntegerDecoder.java
```diff
@@ -1,25 +1,18 @@
...
 import java.math.BigInteger;
-import java.nio.ByteBuffer;
+import java.util.Optional;
 
 public class UnsignedIntegerDecoder<T> extends IntegerDecoder{
     int sign;
 
-    public UnsignedIntegerDecoder(VariableContents contents, int givenSign) {
-        super(contents);
+    public UnsignedIntegerDecoder(int size, int givenSign, String name) {
```
ducky64: What is givenSign? Is this something that can be derived from `size`? If it is, I'd say it's better to keep a minimal interface here (to be correct-by-construction by eliminating potential data inconsistencies).

app/src/main/java/com/example/bottomnav/UnsignedLongDecoder.java
```diff
@@ -1,23 +1,17 @@
 package com.example.bottomnav;
 
 import java.math.BigInteger;
-import java.nio.ByteBuffer;
+import java.util.Optional;
 
 public class UnsignedLongDecoder<T> extends IntegerDecoder {
-    Long sign;
+    Long sign = (long) 0xffffffffL;
```
ducky64: Does this need its own class? Can you make `sign` in UnsignedDecoder a long?

Or better, automatically derive `sign` so it's not part of the interface and have UnsignedIntegerDecoder also support longs natively.

app/src/main/java/com/example/bottomnav/VariableContents.java
```diff
@@ -1,20 +1,121 @@
...
+                    (CPPASTDeclarator) dec.getDeclarators()[0]);
+        } else {
+            return makeOptFromDecl((CPPASTNamedTypeSpecifier) dec.getDeclSpecifier(),
+                    (CPPASTDeclarator) dec.getDeclarators()[0]);
+        }
+    }
+
+    /**
+     *
+     * @param simpleSpec
```
ducky64: If you're going to make the whole javadoc block you should fill out all the parameter documentation too. Or just get rid of the parameters in the javadoc block, or just use a standard comment (`//`) if you're just adding a note instead of javadoc.

app/src/main/java/com/example/bottomnav/VariableContents.java
```diff
@@ -1,20 +1,121 @@
...
+     * @param declarator
+     * @return Optional<Components>
+     * @throws Exception
+     *
+     * Difference: Simple specifiers don't quantifier types and do not have name object
+     */
+    private static Optional<VariableContents> makeOptFromDecl(CPPASTSimpleDeclSpecifier simpleSpec,
+                                                              CPPASTDeclarator declarator) throws Exception {
```
ducky64:
```suggestion
                                                              CPPASTDeclarator declarator) Exception {
```
I don't think this throws exceptions anymore? (and similarly below)

app/src/main/java/com/example/bottomnav/VariableContents.java
```diff
@@ -1,20 +1,121 @@
...
+    /**
+     *
+     * @param simpleSpec
+     * @param declarator
+     * @return Optional<Components>
+     * @throws Exception
+     *
+     * Difference: Simple specifiers don't quantifier types and do not have name object
+     */
+    private static Optional<VariableContents> makeOptFromDecl(CPPASTSimpleDeclSpecifier simpleSpec,
```
ducky64:
```suggestion
    private static Optional<VariableContents> fromDecl(CPPASTSimpleDeclSpecifier simpleSpec,
```
simpler name, especially since there isn't a non-Optional version (and similarly below)

app/src/main/java/com/example/bottomnav/VariableContents.java
```diff
@@ -1,20 +1,121 @@
...
+            String typeQualifierNameStr = namedTypeSpec.getSyntax().getImage();
+            String primitiveTypeNameStr = namedTypeSpec.getName().getRawSignature();
+            return Optional.of(new VariableContents(declaratorNameStr, typeQualifierNameStr,
+                    primitiveTypeNameStr, initOpt));
+        } else {
+            return Optional.empty();
+        }
+    }
+
+    private static Optional<CPPASTEqualsInitializer> makeOptInit(CPPASTEqualsInitializer init) {
```
ducky64: This might be redundant with `Optional.ofNullable(...)`

app/src/main/java/com/example/bottomnav/VariableContents.java
```diff
@@ -1,20 +1,121 @@
...
+        if (init != null) {
+            return Optional.of(init);
+        } else {
+            return Optional.empty();
+        }
+    }
+
+    private Optional<String> getValue(Optional<CPPASTEqualsInitializer> givenInit) {
```
ducky64: This Optional to Optional feels a bit weird (non-orthogonality - the Optional wrapper should be handled by an upper layer).
Also, what is this used for?

app/src/main/java/com/example/bottomnav/VariableContents.java
```diff
@@ -1,20 +1,121 @@
...
+        return Optional.empty();
     }
 
-    public void setPacketSize(int size) {
-        packetSize = size;
+    /**
+     * @param primType
+     * @return char[]
+     *
+     * AST parser retrieves an integer to represent a primitive type. This function allows to
```
ducky64: Is there a function in the parser / AST library that does this? I'd expect (more likely than not, but not 100%) that there would be some utility to turn those numbers into something readable.

If you must do this yourself, consider having an enum so it's more type safe.

app/src/main/java/com/example/bottomnav/ui/notifications/NotificationsFragment.java
```diff
@@ -31,6 +42,73 @@ public View onCreateView(@NonNull LayoutInflater inflater,
             @Override
             public void onChanged(@Nullable String s) {
                 textView.setText(s);
+        adapter = new ArrayAdapter<>(getActivity(),
```
ducky64: why is the indenting super weird?

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -197,6 +89,7 @@ public void decodeIntegers() throws Exception {
 
     }
 
+    /**
```
ducky64: Why'd this test get commented out?

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -3,180 +3,72 @@
...
-        assertEquals("uint16_t", ((IntegerDecoder) test.getDecoder("CAN_ID2").get()).contents.payloadDataType);
-        assertEquals("uint16_t", ((IntegerDecoder) test.getDecoder("CAN_ID3").get()).contents.payloadDataType);
-
-        assertEquals("0x16", ((IntegerDecoder) test.getDecoder("CAN_ID").get()).contents.value);
-        assertEquals("0x32", ((IntegerDecoder) test.getDecoder("CAN_ID2").get()).contents.value);
-        assertEquals("0x64", ((IntegerDecoder) test.getDecoder("CAN_ID3").get()).contents.value);
-    }
+        byte[] eight = {(byte) 0xff};
+        assertEquals("BRUH: 255", test.decodeToString(0x402, eight));
+        assertEquals("YOUNG: -1", test.decodeToString(0x403, eight));
```
ducky64: This looks like a lot of removed tests. Unless those tests are now invalid or redundant, removing tests is generally a bit of a red flag.

ducky64 (changes requested):
This looks a lot better! But I think there's still improvements that can be made - see inline comments.

app/src/main/java/com/example/bottomnav/IntegerDecoder.java
MichaelCampos-eng: That is handled in the PrimitiveDecoder class's public function getRawValue.

app/src/main/java/com/example/bottomnav/DataDecoder.java
MichaelCampos-eng: Okedokey

app/src/main/java/com/example/bottomnav/DataDecoder.java
MichaelCampos-eng: Since CreatePrimitiveDecoder static function is in same class, I wanted to keep createeStructDecoder consistent.

app/src/main/java/com/example/bottomnav/DataDecoder.java
MichaelCampos-eng: From the test cases you provided a while ago, the packet size required to decode was 1. Not sure why

app/src/main/java/com/example/bottomnav/VariableContents.java
MichaelCampos-eng: Consts have a value assigned to them, while struct variables don't. There are some cases where CPPASTEqualsInitializer doesn't exist, so I needed to check for that.

app/src/test/java/com/example/bottomnav/ParseTest.java
MichaelCampos-eng: Yeah, the previous tests were no longer compatible to this new version. Most of the deleted tests were only checking if parser part of the code was doing its job. Making the decoder as stateless as possible came at cost of not storing every detail of a constant or structure such as type qualifier and data type.

app/src/main/java/com/example/bottomnav/DoubleDecoder.java
```diff
@@ -2,20 +2,16 @@
...
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
+import java.util.Optional;
 
 public class DoubleDecoder<T> extends PrimitiveDecoder{
 
-    public DoubleDecoder(VariableContents contents) {
-        super(contents);
+    public DoubleDecoder(int size, String name) {
+        super(size, name);
```
ducky64:
```suggestion
    public DoubleDecoder(String name) {
        super(8, name);
```

Double (and Float) have standardized sizes, so it shouldn't take size as an argument.

When you're calling a super constructor, you can pass a constant value (or any value), it doesn't necessarily need to come from a containing constructor argument.

app/src/main/java/com/example/bottomnav/MainActivity.java
ducky64: Can you delete things instead of commenting them out?

Commenting things out is fine for testing and development, but for maintaining a clean codebase committing commented out code is discouraged.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -20,8 +20,7 @@
 
 public class Parse {
     private HashMap<String, Optional<DataDecoder>> decoderRepo= new HashMap<>();
-    private HashMap<String, String> canNameToStruct = new HashMap<>();
-    private HashMap<Integer, String> canIdToName = new HashMap<>();
+    private HashMap<Integer, String> IdToName = new HashMap<>();
```
ducky64:
```suggestion
    private HashMap<Integer, String> idToName = new HashMap<>();
```

Consistent variable naming style?

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -32,6 +31,7 @@ public Parse(char[] code) throws Exception {
         IASTTranslationUnit translationUnit = getIASTTranslationUnit(code);
         IASTNode[] comments = translationUnit.getComments();
         IASTNode[] children = translationUnit.getChildren();
+        HashMap<String, Integer> NameToId = new HashMap<>();
```
ducky64:
```suggestion
        HashMap<String, Integer> NameToId = new HashMap<>();
```

similarly, consistent variable naming style?

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -51,75 +51,79 @@ public Parse(char[] code) throws Exception {
...
+                decoderRepo.put(canName, decoderRepo.get(structName));
             }
         }
     }
 
+    /**
+     * Creates and stores a Primitive decoder if declaration is a Const
+     * @param declaration
+     * @throws Exception
+     */
```
ducky64: Does this actually throw an Exception?

I'm also not sure that `throws Exception` is too useful, if there's a specific exception this might throw (and the caller needs to be aware of) that might be more useful to document.

app/src/main/java/com/example/bottomnav/DataDecoder.java
ducky64: Where was this?

It's also worth keeping in mind that the test cases may be wrong. I'm pretty sure `int` is 4 bytes signed.

app/src/main/java/com/example/bottomnav/VariableContents.java
ducky64: I think what I mean here is that you should structure this function as CPPASTEqualsInitializer -> String, and deal with the Optional separately.

Take a look at `Optional.map`: https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#map-java.util.function.Function-

Essentially, instead of duplicating the function of `Optional.map` in this function, I'd recommend writing the function to do one thing, then compose it with `Optional.map` with the data you're working is wrapped in an Optional.

app/src/main/java/com/example/bottomnav/IntegerDecoder.java
ducky64: Oh, I see, so the precondition to calling this is that the payload will be of the correct size?

I'd recommend documenting these preconditions. It's not immediately obvious just by looking at the code.

ducky64 (changes requested):
See updated comments.

Also, I think a bunch of comments from the prior reviews were not addressed. GitHub tends to do the wrong thing and hides conversations that are not stale or resolved - if it says "show hidden conversations" make sure to expand and read those.

app/src/main/java/com/example/bottomnav/UnsignedIntegerDecoder.java
MichaelCampos-eng: "givenSign" is used for bitwise operation when trying to decode the payload as unsigned. I can't really use a universal hexadecimal number because doing bitwise op. needs both the payload and "givenSign" to be the same length.

app/src/main/java/com/example/bottomnav/MainActivity.java
```diff
@@ -1,17 +1,23 @@
+/**
```
ducky64: Can these be deleted instead of commented out? (actually, looks like the whole file is commented out)

Commented out code is generally discouraged in a clean codebase, especially if it's not well documented why it's commented out - so the next person looking at this gets confused.

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -19,15 +20,16 @@
 import java.util.regex.Pattern;
 
 public class Parse {
-    private HashMap<String, Optional<DataDecoder>> decoderRepo= new HashMap<>();
-    private HashMap<String, String> canNameToStruct = new HashMap<>();
-    private HashMap<Integer, String> canIdToName = new HashMap<>();
+    private HashMap<String, DataDecoder> decoderRepo= new HashMap<>();
```
ducky64:
```suggestion
    private HashMap<String, DataDecoder> decoderRepo = new HashMap<>();
```
minor style nit

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -51,75 +53,167 @@ public Parse(char[] code) throws Exception {
...
+        /**
+         * Declaration specifiers vary!
+         * Type Qualifer: eg. const, volatile, restrict, etc.
+         * Unable to correctly obtain type qualifiers, type qualifiers sometimes to specifier if
+         * qualifer doesn't exist
+         */
+        if (typeQualiferStr == primitiveStr) {
+            typeQualiferStr = null;
         }
+        return Optional.of(new VariableContents(declaratorStr, typeQualiferStr, primitiveStr, valueStr));
```
ducky64: If this function can only return Optional.of (and never Optional.empty), does the return type need to be optional?
Or alternatively, which might be a better solution, when the inputs fail validation (eg, there's an null somewhere), does it make sense to return Optional.empty instead of return an Optional.of(VariableContents), where the VariableContents may be invalid?

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -51,75 +53,167 @@ public Parse(char[] code) throws Exception {
...
+         * Type Qualifer: eg. const, volatile, restrict, etc.
+         */
+        if (declaration.getDeclSpecifier() instanceof CPPASTSimpleDeclSpecifier) {
+            CPPASTSimpleDeclSpecifier simpleSpecifier =
+                    (CPPASTSimpleDeclSpecifier) declaration.getDeclSpecifier();
+            if (declarator != null && simpleSpecifier != null) {
+                typeQualiferStr = simpleSpecifier.getSyntax().getImage();
+                primitiveStr = getPrimitiveType(simpleSpecifier.getType());
+            }
+        } else {
```
ducky64: Do you need to check for `instanceof CPPASTNamedTypeSpecifier`?

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -51,75 +53,167 @@ public Parse(char[] code) throws Exception {
...
-            Optional<Components> data = Components.getComponents(declaration);
-            if (data.isPresent() && !data.get().init.isPresent()) {
-                Components comp = data.get();
-                VariableContents contents = new VariableContents(comp.name, null, null, comp.type);
-                variables.add(contents);
-            }
-        }
-        Optional<DataDecoder> decoder = Optional.of((DataDecoder) new StructDecoder(variables));
-        decoderRepo.put(name, decoder);
+    private VariableContents isAppropriateConst(VariableContents contents){
```
ducky64: Instead of having nulls around, does it make sense to have this be a function in VariableContents, eg VariableContents.isConst?

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -51,75 +53,167 @@ public Parse(char[] code) throws Exception {
...
+            return contents;
+        } return null;
     }
 
-    public String decode(int canId, byte[] payload) {
-        String canName = canIdToName.get(canId);
-        Optional<DataDecoder> decoder = getDecoder(canName);
-        if (decoder.isPresent()) {
-            return decoder.get().decode(canId, payload);
+    private VariableContents isAppropriateMember(VariableContents contents){
```
ducky64: Similarly here, should this be moved into VariableContents?

app/src/main/java/com/example/bottomnav/Parse.java
```diff
@@ -51,75 +53,167 @@ public Parse(char[] code) throws Exception {
...
+    }
+
+    /**
+     * The decoders return a string format of 'NAME: Value, NAME: Value, ...'
+     * Therefore, parseStringResults returns an array of Decoded objects that allows user to iterate
+     * over and use functions like getName and getValue in for each Decoded object
+     *
+     * @return
+     */
+    public ArrayList<DecodedContents> parseDecodedString(String resultOfDecoder) {
```
ducky64: Packing structured data into a string then reverse-engineering that string into structured data is kind of a red flag. Can this be done more directly?

app/src/main/java/com/example/bottomnav/StructDecoder.java
```diff
@@ -1,47 +1,78 @@
...
+    public Optional<String> decodeToString(byte[] payload) {
+        ArrayList<String> decodedStrings = new ArrayList<>();
+        for(int index = 0; index < decodedPrimitives.size(); index++) {
+            PrimitiveDecoder decoder = (PrimitiveDecoder) decodedPrimitives.get(index);
+            byte[] dedicatedPayload = getDedicatedPayload(payload, decoder.typeSize);
+            if (index == decodedPrimitives.size() - 1) {
+                if (dedicatedPayload.length < payload.length) {
+                    decodedStrings.add("incorrectlyFormattedPayload: null");
+                    break;
+                }
```
ducky64: Can you structure this more like the above?
`for (DataDecoder decoder : decodedPrimitives) { ...`

I know that on the last element you do a payload length check, but do keep in mind that you could get an unexpectedly short message at any index, for example perhaps the message had an empty payload

app/src/main/java/com/example/bottomnav/StructDecoder.java
```diff
@@ -1,47 +1,78 @@
...
+            } else {
+                break;
+            }
+            payload = adjustPayload(payload, decoder.typeSize);
+        }
+        return Optional.of(String.join(",", decodedStrings));
+    }
+
+    // Splice function that return an array of bytes under specified
+    private byte[] getDedicatedPayload(byte[] payload, int typeSize){
```
ducky64: Would it make sense to have this return `Optional<byte[]>` instead of returning a truncated array?
This gives you stronger guarantees that if this returns a value, it will be exactly typeSize long

app/src/main/java/com/example/bottomnav/UnsignedIntegerDecoder.java
ducky64: I don't think you can use a universal hex number, but you could derive the hex number from the byte size right?
For example, size=1 gives `0xff`, size=2 gives `0xffff`, size=4 (invalid here) gives `0xffffffff`, and in general, does `givenSign=1 << (8 * size) - 1` work? (note, `1 << (8 * size)` is the bit shift version of 2 ^ (8 * size))

app/src/main/java/com/example/bottomnav/UnsignedLongDecoder.java
ducky64: Are you sure you need a separate UnsignedLongDecoder? This is a superset of UnsignedIntegerDecoder, but a bit less (inconsequentially) memory efficient since it needs to store a long instead of an int? But the trade-off for this efficiency is that you now have more code the maintain.

In fact, I'd even say that given the expected use case, if we used a BigInteger internally to unify all the int-like decoders that would still be worth the code maintenance tradeoff.

app/src/main/java/com/example/bottomnav/ui/notifications/NotificationsFragment.java
```diff
@@ -1,5 +1,5 @@
 package com.example.bottomnav.ui.notifications;
-
+/**
```
ducky64: Can you delete the commented out code?

app/src/test/resources/structs.h
```diff
@@ -0,0 +1,23 @@
...
+//
+
+#ifndef BOTTOM_NAV_STRUCTS_H
+#define BOTTOM_NAV_STRUCTS_H
+
+
+// @canPayloadStruct CAN_PEDAL_POS = CanPedalPosStruct
+const uint16_t CAN_PEDAL_POS = 0x282;
+
+struct CanPedalPosStruct {
```
ducky64, line 11: fyi this actually won't fit within CAN, but it's fine as a test case
CAN payloads are 8 bytes max, this is 4+1+1+2+2+4+4=18

ducky64 (changes requested):
Looking better, almost there! A few more suggestions, which hopefully should just be localized changes, and I think it's good to go.

There's a lot of commented-out code (as mentioned in the inline comments), if it's dead code can you just delete it?

app/src/main/java/com/example/bottomnav/Parse.java
MichaelCampos-eng: Yes. CPPASTSImpleTypeSpecifier only allows me to access the var/member specifier by calling getType(), but it returns an integer corresponding to a type specifier. On the other hand, CPPASTNamedTypeSpecifier does have the same function getType(), but only returns 1, probably meaning that the specifier type is "custom." But CPPASTNamedTypeSpecifier has a function called getName() which return the actual name of the type specifier in string.

app/src/main/java/com/example/bottomnav/StructDecoder.java
MichaelCampos-eng: Incorrect byte size is checked when a DecoderPrimitive decodes something.

app/src/main/java/com/example/bottomnav/Parse.java
MichaelCampos-eng: I believe that the decoders themselves should only return an optional of a string.  I thought about returning an optional of an object where variables would be name and value.  But it gets a little complicated because struct decoder uses primitive decoder since I want them to work universally: Decode(payload) => optional(string)

app/src/main/java/com/example/bottomnav/DecoderStruct.java
```diff
@@ -43,25 +43,35 @@ public Optional<ArrayList<Optional>> decodeToRaw(byte[] payload) {
...
-                if (dedicatedPayload.length < payload.length) {
-                    decodedStrings.add("incorrectlyFormattedPayload: null");
-                    break;
-                }
-            }
-            Optional<String> decoded = decoder.decodeToString(dedicatedPayload);
+        ArrayList<String> decodedContents = new ArrayList<>();
+
+        for(DecoderData decoder : decodedPrimitives) {
+            DecoderPrimitive decoderPrimitive = (DecoderPrimitive) decoder;
```
ducky64: Is this guaranteed? Because it's possible to have nested structs, right?

If it's guaranteed, I'd recommend changing `decodedPrimitives` to `ArrayList<DecoderPrimitive>`. If not, consider just adding a comment along the lines of `// TODO support nested structs`

app/src/main/java/com/example/bottomnav/Statement.java
```diff
@@ -0,0 +1,116 @@
...
+        if (decl != null) {
+            declarator = Optional.of(decl);
+        } else {
+            declarator = Optional.empty();
+        }
+        if (init != null) {
+            initializer = Optional.of(init);
+        } else {
+            initializer = Optional.empty();
+        }
```
ducky64:
```suggestion
        declarator = Optional.ofNullable(decl);
        initializer = Optional.ofNullable(init);
```
Check out Optional.ofNullable!

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -3,30 +3,40 @@
...
         assertEquals(false, second.isPresent());
     }
 
+    @Test
+    public void testing() throws Exception {
+        Parse test = Parse.parseTextFile("constants.h");
+        byte[] eight = {(byte) 0xff};
+
+        Optional<DecoderData> decoder = test.getDecoder(0x402);
+        System.out.println(decoder.get().isPrimitive());
```
ducky64: Stray debugging `println`?

Consider replacing with an assertion.

app/src/main/java/com/example/bottomnav/Statement.java
```diff
@@ -0,0 +1,116 @@
...
+            simpleSpecifier = Optional.empty();
+            namedTypeSpecifier = Optional.of((CPPASTNamedTypeSpecifier) declaration.getDeclSpecifier());
+        } else { // Possibility that specifier is neither simple or named type.
+            simpleSpecifier = Optional.empty();
+            namedTypeSpecifier = Optional.empty();
+        }
+    }
+
+    // Checks that both declarator and specifier exist. All variables and members must have them.
+    protected boolean isStatementValid() {
```
ducky64: I'd prefer ways to make it so that if you have an instance, it's valid - instead of getting an objecting and then needing to test for validity. We might have to think a bit about how to make this structurally work.

ducky64 (changes requested):
Mostly looks good, I think the only issue left is with parseDecoded and it being possible to construct an invalid Statement object. Perhaps we should find some time to think about ways to make sure it's never possible to get an invalid Statement object?

Did you also want to combine UnsignedIntegerDecoder with UnsignedLongDecoder?

app/src/main/java/com/example/bottomnav/Parse.java
ducky64: parseDecoded looks like it's only used in test cases? If that's the case, consider moving it to the test code as a test helper function.

In general, I'd say that structured data -> string -> decode back into structured data is a major red flag and can be a longer term software maintenance headache.

app/src/main/java/com/example/bottomnav/DecoderStruct.java
```diff
@@ -0,0 +1,83 @@
...
+import java.lang.*;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.Optional;
+
+public class DecoderStruct extends DecoderData {
+    private ArrayList<DecoderData> decoders;
+    private int typeSize;
+
+    static Optional<DecoderData> create(int numMembers, int byteSize, ArrayList<DecoderData> decoders) {
```
ducky64, line 12: Any reason this also takes members-count and decoders separately? It seems to be that this consistency check (if needed?) should be handled at the caller level, unless there's something struct-specific about this?

app/src/main/java/com/example/bottomnav/DecoderStruct.java
```diff
@@ -0,0 +1,83 @@
...
+    @Override
+    public int getTypeSize() {
+        return typeSize;
+    }
+
+    @Override
+    public Optional<ArrayList<DecodedData>> decodeToRaw(byte[] payload) {
+        ArrayList<DecodedData> decodedArray = new ArrayList<>();
+        for(DecoderData decoder : decoders) {
+            byte[] dedicatedPayload = getDedicatedPayload(payload, decoder.getTypeSize());
```
ducky64, line 33: Would it be simpler to check that the payload is >= to the expected size, and if not just return Optional.empty?
getDedicatedPayload is doing a lot of work to clean up invalid data that could just be presented as invalid?

app/src/test/java/com/example/bottomnav/ParseTest.java
```diff
@@ -1,113 +1,153 @@
...
+        byte[] payload = {(byte) 0xff, (byte) 0xff, (byte) 0xff, (byte) 0xff, (byte) 0xff,
+                (byte) 0xff, (byte) 0xff, (byte) 0xff, (byte) 0xff, (byte) 0xff, (byte) 0xff,
+                (byte) 0xff, (byte) 0xff, (byte) 0xff, (byte) 0xff, (byte) 0xff};
+        DecodedData structDecoded = test.decode(0x300, payload).get();
+        assertEquals(true, structDecoded instanceof DecodedStruct);
+        ArrayList<DecodedData> members = ((DecodedStruct) structDecoded).getMembers();
+        assertEquals(true, members.get(0) instanceof DecodedPrimitive);
+        assertEquals(true, members.get(1) instanceof DecodedStruct);
+        assertEquals(true, members.get(2) instanceof DecodedStruct);
+        assertEquals("positionDegree: -1, impulseResponse: -1, frequencyResponse: -1, pascal: -1", structDecoded.dataToString());
```
ducky64, line 148: Perhaps a future TODO: see if you can preserve the nested structure instead of flattening everything

ducky64 (approved):
This looks fine, and I think at this point we should prioritize getting this merged and unblocking further work rather than getting it perfect.

There's a few inline comments with suggestions for minor improvements, but fixes are not required.

I still think it would be cleaner if there was a way to make it impossible to have an invalid Statement - it's definitely structurally tricky and I understand the compromise you made. We can discuss this if you'd like. But I'd highly prefer that be a separate refactoring PR, both to get this merged to unblock other work, and to have smaller and more focused PRs.

